PR #2, opened 2017-07-25
0002 - Extracting Necko.

rnewman:


joewalker:
I agree with this, but I think people with a better knowledge of Gecko should weigh in. @petervanderbeken / @mykmelez

joewalker (commented):
See [earlier comment](https://github.com/mozilla/firefox-browser-architecture/pull/2#issuecomment-318416204)

rnewman:
Rebased on #3.

README.md
```diff
@@ -15,7 +14,8 @@ Our top priority is to plan for where we want to go, but if we just think 5 year
 ## Our Conclusions
 
 * [Documenting our output](text/0001-documenting-output.md) looks at how we’re going to communicate with the rest of Mozilla.
+* [Extracting Necko](text/0002-extracting-necko.md) considers whether it's feasible or worthwhile to extract Necko — Gecko's C++ networking library — for use as a standalone component.
 
 ## Posts
 
-* [Browser Architecture Update](posts/2017-07-27-browser-architecture-update.md) is the first of our newsletters, sent to firefox-dev to keep everyone in the loop.
+* [Browser Architecture Update](posts/2017-07-27-browser-architecture-update.md) is the first of our newsletters, sent to [firefox-dev](https://mail.mozilla.org/pipermail/firefox-dev/) to keep everyone in the loop.
```
mykmelez: For a link to the "firefox-dev" discussion forum generally, I would point to either its entry on the mozilla.org Forums page:

https://www.mozilla.org/en-US/about/forums/#firefox-dev

Or its Wikimo page:

https://wiki.mozilla.org/Firefox/firefox-dev

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,67 @@
...
+
+# References
+
+Lots of very old documents:
+
+* <https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Necko/Embedding_Necko>
+* <https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Necko>
+* <https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Necko/Architecture> 
+* Discussions with rhelmer and jduell
+* Examination of `/netwerk`
```
mykmelez: The formatting here suggests that "discussions with rhelmer and jduell" and "examination of `/netwerk`" are both "very old documents." I think you meant something like this instead:

* lots of very old documents
** very old document a
** very old document b
** very old document c
* discussions with rhelmer and jduell
* examination of `/netwerk`

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,67 @@
...
+* Parts of Necko are still single-threaded, requiring new channels to be created on the main thread, and dispatching callbacks to the main thread.
+* Necko's init and deinit are managed by observer notifications, which is not ideal for embedding outside of Gecko.
+
+# Proposal
+
+Necko is a very feature-rich C++ networking library that is directly targeted at meeting the needs of single-user, single-main-process, single-profile, interactive applications built on a persistent profile directory in the traditional Mozilla model. It is not well-suited for use outside of the Gecko lifecycle — indeed, almost every aspect is designed for the Gecko world — and would require substantial effort in decoupling to be suited for embedded use. Even if so decoupled, it would still mandate the use of XPCOM and the NSPR.
+
+To extract Necko for reuse outside of Gecko, by non-XPCOM-based applications, would — almost by definition — require coming up with a plan to deal with XPCOM. Given that Necko's data structures, interface definitions, macros, etc. *are* XPCOM's, it seems infeasible to eliminate XPCOM dependency from the module without a total rewrite.
```
mykmelez: The first assertion here—that extracting Necko requires a plan for XPCOM—seems reasonable to me. The second one—that it's infeasible to eliminate XPCOM without a total rewrite—is less obvious.

It may be hard to know for sure without knowing the plan (specifically: what replaces XPCOM). But at least for the XPIDL interfaces, assuming the alternative to XPCOM is something like a native C++ class-based interface, then it seems like it would be feasible to replace XPIDL by landing the C++ header files that the XPIDL compiler generates and then evolving them into the desired interface without rewriting much of their implementations.

However, I'm far from an expert on XPCOM, and someone who is should provide input here.

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,67 @@
...
+
+This is particularly relevant when we consider oxidation — the replacement of C++ and JS components of Firefox with Rust — and even more so when we consider the use of such components both outside and inside Firefox *from* Rust (rather than from existing non-Rust callers).
```
mykmelez: Oxidation needs to solve the Rust FFI problems more generally, perhaps including Rust implementations of XPCOM components (in addition to Rust crates accessing XPCOM components written in C++ and JS), so presumably the existing, integrated (into Gecko) implementation of Necko could reuse those solutions.

A standalone Necko (for reuse by Rust crates) seems easier, since it can ditch XPCOM and use whatever the best practice is for Rust access to C++ code (unsure what that is, however). Nevertheless, the difficulty then becomes how to reuse the standalone Necko in Gecko.

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,67 @@
...
+
+Our position is that extracting Necko from Gecko for reuse would be a significant amount of work, and doing so in a way that allows its use from ordinary Rust consumers, without shipping XPCOM/NSPR/*etc*., would be even more so.
+
+Our proposal, then, is that we should not attempt to reuse Necko outside of Gecko.
```
mykmelez: The argument presented over the last two paragraphs sounds like:

1. It would be hard.
2. Therefore we should not do it.

That misses a key step, which is that extracting Necko doesn't provide enough value to justify the work. After all, if it was hard, but it was worth it, then we'd want to do it anyway (c.f. GeckoView, a UA service, etc.).

So I would add a paragraph between "Our position is…" and "Our proposal, then…" that describes the extent of the value we expect to gain from extracting Necko, and how it's not commensurate with the work involved.

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,67 @@
+# Problem
+
+Can we break out a Gecko component, such as Necko, for reuse in background services and on mobile platforms?
+
+## Motivation
+
+If — as we expect — Mozilla will need to build consistent native applications across platforms, and background services on desktop, it would be beneficial in theory to leverage our existing investment in our network layer, avoiding the confusing situation of having two ways to configure network settings, and avoiding the possibility of different behaviors in different parts of an application.
```
mykmelez: Don't we want to build background services on mobile as well? Or does this assume that background services on mobile will use mobile platform-specific equivalents to Necko?

mykmelez (changes requested):

text/0002-extracting-necko.md
rnewman: Yes, we do. The mention of desktop background services in particular is an attempt to imply _without Gecko and a profile running_, which is the desktop equivalent to native mobile apps that don't have Gecko at all. I'll clarify.

text/0002-extracting-necko.md
rnewman: More input is very welcome!

When I read through the Necko source code, my impression was that practically every method and class somewhere used services, `ns*` classes and data structures, XPCOM errors and results and macros, *et al*. And so my impression is that eliminating what we would call "XPCOM", either narrowly (build and link without use of `/xpcom/`) or broadly (no non-standard C++, including no NSPR and no mozglue and no mfbt), would involve touching almost every line of code.

There are certainly options here — *e.g., * [we could resurrect Oink and use it to rewrite the C++ to something modern](https://brendaneich.com/2006/10/mozilla-2/) — but I still consider that to be a rewrite.

What I consider _not_ to be a total rewrite is being able to grab whole files out of Necko and reuse them with minimal changes: that is, if Necko already carefully encapsulated its use of `nsTArray` and `NS_SUCCEEDED` and `nsCString` and friends, so whole subcomponents were already safe. That doesn't seem to be the case (but I'd be happy to be proved wrong).

Do you think it would be worth splitting this argument into deCOMtamination and "other changes"?

text/0002-extracting-necko.md
rnewman: Yes, exactly. Necko seems particularly tricky here: some components have a very narrow set of consumers, and so one can oxidize up to an existing IPC or well-contained boundary, but Necko channels permeate the codebase.

text/0002-extracting-necko.md
rnewman: I'll probably do this by cutting up some of the next paragraph — "Rust components should use Rust networking libraries of equivalent functionality" — to turn the implication (minimal value because we already have Hyper and friends) into an explication.

rnewman:
Thanks for the thorough review, @mykmelez!

text/0002-extracting-necko.md
mykmelez: I suppose I think of a "total rewrite" as a from-scratch reimplementation that conserves—at most—only the API (and possibly tests).

Thus a "rewrite" that conserves the existing implementation while eliminating XPCOM from it isn't a "total rewrite," even if it ends up touching every line of code. I'm not sure I would even use the word "rewrite" in that case, as it seems more like a "refactoring." And it might even be the kind of work that can proceed in (carefully planned) stages.

The example in Brendan's blog is illustrative, as it demonstrates how the removal of XPCOM (and other archaisms) results in changes on almost every line while still conserving the existing implementation. Here's a diff of his before and after blocks:

```diff
-PRBool
-nsXULDocument::OnDocumentParserError()
+bool
+XULDocument::OnDocumentParserError()
 {
     // don't report errors that are from overlays
     if (mCurrentPrototype && mMasterPrototype != mCurrentPrototype) {
-        nsCOMPtr<nsIURI> uri;
-        nsresult rv = mCurrentPrototype->GetURI(getter_AddRefs(uri));
-        if (NS_SUCCEEDED(rv)) {
-            PRBool isChrome = IsChromeURI(uri);
-            if (isChrome) {
-                nsCOMPtr os(
-                do_GetService("@mozilla.org/observer-service;1"));
-                if (os)
-                    os->NotifyObservers(uri, "xul-overlay-parsererror",
-                EmptyString().get());
-            }
+        IURI *uri = mCurrentPrototype->GetURI();
+        if (IsChromeURI(uri)) {
+            GetObserverService()->NotifyObservers(uri, "xul-overlay-parsererror");
         }
-        return PR_FALSE;
+        return false;
     }
-    return PR_TRUE;
+    return true;
 }
```

The only lines that aren't touched are a line containing a comment, another containing a conditional, and a few containing opening/closing braces. Most lines look very different. But the implementation itself is still conserved. This function still does exactly the same thing.

mykmelez: I think I would differentiate between XPCOM and NSPR/MFBT (unsure about mozglue). The distinction is that XPCOM affects not only the internal implementation but also the external interface, while NSPR/MFBT do not.

That may be presumptuous, as MFBT (at least) defines alternatives to standard C++ features like smart pointers, and those alternatives may leak into the interface. But it still seems more plausible to plug those leaks while continuing to use MFBT in the implementation. Here again I'd love someone with more experience to validate my impression.

> Do you think it would be worth splitting this argument into deCOMtamination and "other changes"?

Yes, I think that'd make sense. In particular, it'd useful to differentiate between the changes required to expose a usable interface to Rust code and those required to "modernize" Necko's internal implementation. (Even if it doesn't necessarily change the answer, which also depends on how useful it would be to do even the smaller subset of work.)

text/0002-extracting-necko.md
mykmelez: That sounds like a good plan!

mykmelez (approved):

text/0002-extracting-necko.md
```diff
@@ -0,0 +1,74 @@
...
+
+Extracting such a module is not routine work: delicate balances need to be found between maintaining useful integration with Gecko (*e.g.*, Necko's `nsIChannel` permeates the codebase) versus designing for embedding, and between reimplementing Gecko dependencies in Rust for reuse versus reworking the component to avoid them.
+
+Our position is that extracting Necko from Gecko for reuse would be a significant amount of work. Doing so in a way that allows its use from ordinary Rust consumers, without shipping XPCOM/NSPR/*etc*., would be even more so.
+
+That amount of work is certainly much more than simply using the existing well-tested and idiomatic networking libraries available to Rust code. Given that the benefits of using Necko in its entirety (instead of *e.g.*, just using Hyper for HTTP) are relatively small, our proposal is that we **should not attempt to reuse Necko outside of Gecko**.
```
mykmelez: Super-pedantic nit: "we should not" sounds more like an assertion or conclusion than a proposal.

So I would rephrase to either "our proposal is that we **not attempt to reuse Necko outside of Gecko**" or "our conclusion is that we **should not attempt to reuse Necko outside of Gecko**".

mykmelez:
Last note: it'll be interesting to see if a conclusion like this remains true over time. It may be that reuse of an XPCOM component in Rust applications becomes easier, or perhaps Gecko will undergo further rounds of deCOMtamination that makes its components more reusable in Rust applications.

Unsure if that's worth calling out explicitly, however, as it's implicitly true for all conclusions we publish here: they are products of their time, always potentially subject to reconsideration. The important question here is whether it makes sense to extract Necko *now*, and that question has been answered.

text/0002-extracting-necko.md
rnewman: I think it's important to capture our _judgment_ here, so I think your second suggestion is best. However, I don't think these documents should be passive: we're not just concluding, we're also proposing! That's probably adequately addressed by the heading of the section, but I'm curious whether you agree.

mykmelez: Yes, I think it's adequately addressed!

In the future, I might hoist the conclusion to the top of the Proposal section for documents like this one, to make it easier to find (since currently most of the text in the Proposal section is reasoning, and there's a lot of it, which means you have to read/skim for a while to get to the conclusion).

I might also distinguish between the proposal (which for this document is effectively "extract Necko") and the conclusion (in this case "we should not extract Necko after all").

floatingatoll: Hi. I write a lot of proposals these days. As an outsider to your (new) process here, I would offer an alternate terminology for this for y'alls consideration:

The "proposal" is "we should NOT extract Necko", so your document structure would be:

- proposal ('do not extract necko')
- supporting evidence ('XPCOM argh')
- paths not advised by the proposal ('extract necko', 'port XPCOM', etc).

And then the proposal is, literally, 'proposed' to the person who has final say. They either accepts its recommendation as the final decision, or negotiate/decree a compromise, or negotiate/decree a choice other than the one proposed.

This does imply that the proposal does not necessarily reflect the final decision unless you go back and add a "Final Decision" section somewhere. That helps avoid the "retrofit proposal to the final decision" problem, while still allowing the proposal itself to contain the final result.

Food for thought!

mykmelez: Thanks for the feedback! I agree that we should avoid retrofitting the proposal to the final decision. The proposal is the assertion that initiates the research/analysis, and it's worth retaining it in its original form, even if the research/analysis winds up disproving it, so that readers understand the genesis of the process.

The final decision, likewise, might differ from the conclusion, so it's worth distinguishing between them as well. To provide an example from another domain: a drug company might propose regulatory approval of a drug (the proposal); an expert panel might then conclude that the drug should not be approved (the conclusion); and an administrator might then override that conclusion and approve the drug (the final decision).

In this case, the proposal was actually to "extract Necko," as I understand it (perhaps from discussions with @joewalker, although I don't remember exactly). @rnewman then did some research/analysis and drawn the conclusion that we should not do so. (It isn't clear that someone in a position of authority has made a "final decision" in this case, or even if it's necessary to do so.)

Reviewing the document, however, it isn't necessarily clear that "extract Necko" is the proposal and "we should not extract Gecko" is the conclusion, since the document starts with a Problem section that asks "Can we break out a Gecko component, such as Necko…?" and then continues with a Proposal section that states the conclusion "we should not attempt to reuse Necko outside of Gecko."

So I would revise it to start with a Proposal section that reads as a statement like "We should extract Necko for reuse in background services and on mobile platforms" and then rename the Proposal section to Conclusion. I'll submit a pull request to that effect, where we can discuss structural improvements further!

mykmelez: I've now submitted pull request #10 for this.

PR #4, opened 2017-08-04
Create 0002-problems-with-xul.md

Mossop:
Add a document enumerating the various problems we have heard about XUL


Mossop:
Can you review @joewalker?

text/0002-problems-with-xul.md
```diff
@@ -0,0 +1,69 @@
...
+## No new features
+
+There is no new feature implementation going on in XUL compared to HTML which
+receives constant improvements to meet the demands of the web.
+
+## Performance issues
+
+It is thought that XUL’s performance is worse than that of HTML where we have spent
+large amounts of time optimising for the web but little amount of time optimising
+for XUL.
```
joewalker: There is a counter argument to this which we should perhaps acknowledge; XUL layout is designed for user interfaces where elements do not generally flow like text does. HTML flexbox (a natural replacement for XUL layout) _is_ designed for document layout, and therefore the algorithms required to implement HTML flexbox are considerably more compute intensive than the XUL equivalent, which can lead to cases where XUL is inherently faster even without optimization.

Also a counter-counter  argument; It is possible that grid layout is a better alternative than flexbox, with better performance characteristics. I'm not sure of the detail.

Either way I don't think this nullifies the performance point.

joewalker (approved):
For me all the points in your document could be summarized as "We don't care enough".

There is an easy counter argument - "Start to care more".

However the reason we don't care enough is structural. Effort towards improving XML-ish user interface technology very quickly gets taken by HTML, which is what we really care about. Mozilla sees advancing HTML as part of it's core mission.

So the "Start to care more" argument is ultimately doomed while that structural inequality remains.

Thanks for this. I'm r+, but I'd like for @jesup and @petervanderbeken to have a chance to reply before we merge. Could you wait a couple of days please.

joewalker:
One other thing - there is a race for 0002 between this PR and https://github.com/mozilla/firefox-browser-architecture/pull/2. Something to resolve before merging.

joewalker:
So the other 0002 landed first, so this should be 0003. The good news is that Brian went for 0004.

Separately it occurs to me that we need a problems-with-xbl.md document too. Care to do that?

Mossop:
Renamed

PR #5, opened 2017-08-07
Create 0004-xbl-web-components.md

bgrins:
Not ready for merge yet -  I'm still waiting on results from a test run about the usage of XBL in the Firefox frontend

### DIFF
--- a/0004-xbl-web-components.md
+++ b/0004-xbl-web-components.md
@@ -34,9 +34,7 @@ helpers: {
   "focus|capturing": () => { /* copy in xbl implementation */ }
 }
 ```
-7. `[inherits]`: This is a way to declaratively copy attributes from a binding parent to content and keep them in sync, similar to the `<observes>` element. For example, this copies the feedURL from the parent into the 'value' field of the label, and also observes future changes
-`<feed feedURL="http://foo.com">
-<xul:label xbl:inherits="value=feedURL,tooltiptext=feedURL">`
+7. `[inherits]`: Is a way to declaratively copy attributes from a binding parent to content and keep them in sync, similar to the `<observes>` element. For example, this copies the feedURL from the parent into the 'value' field of the label and also observes future changes: `<feed feedURL="http://foo.com"><xul:label xbl:inherits="value=feedURL,tooltiptext=feedURL"></feed>`. The same thing could be built on top of Custom Elements using a MutationObserver and some glue code to set it up in the `connectedCallback`.
 8. Scoped CSS Styling: this can be done with Shadow DOM
 9. Insertion Points: XBL and Shadow DOM appear to use similar models but different keywords. Further investigation is needed.
 
--- a/0004-xbl-web-components.md
+++ b/0004-xbl-web-components.md
@@ -39,7 +39,6 @@ helpers: {
 <xul:label xbl:inherits="value=feedURL,tooltiptext=feedURL">`
 8. Scoped CSS Styling: this can be done with Shadow DOM
 9. Insertion Points: XBL and Shadow DOM appear to use similar models but different keywords. Further investigation is needed.
-10. `[role]`: Used for accessibility: `role="xul:toolbar"`
 
 ### XBL features that don't map Web Components features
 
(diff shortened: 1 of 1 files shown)


bgrins:
Current element list is at: https://bug1376546.bmoattachments.org/attachment.cgi?id=8896023 / https://bugzilla.mozilla.org/show_bug.cgi?id=1376546. I wouldn't consider it exactly right, but it gives an approximation of which bindings are most used in the browser chrome.

0004-xbl-web-components.md
```diff
@@ -0,0 +1,75 @@
...
+4. `<destructor>`: These map to `disconnectedCallback` on the Custom Element class
+5. `[extends]`: A way to extend the functionality of another binding, i.e. `extends="chrome://global/content/bindings/general.xml#basetext"`. On a surface level this appears to map directly to how the ‘extends’ keyword in ES6 classes behavior with Custom Elements, but more investigation is needed.  There’s a small set of tags where you can do `extends="xul:button"` but that’s aliasing the same thing.
+6. `<handlers>`: Can be implemented with addEventListener or possibly a helper on the base element i.e:
+```
+helpers: {
+  "click": () => { /* bubble by default*/ }
+  "focus|capturing": () => { /* copy in xbl implementation */ }
+}
+```
+7. `[inherits]`: This is a way to declaratively copy attributes from a binding parent to content and keep them in sync, similar to the `<observes>` element. For example, this copies the feedURL from the parent into the 'value' field of the label, and also observes future changes
```
Mossop: You don't say how this can be implemented in web components. Can it?

0004-xbl-web-components.md
```diff
@@ -0,0 +1,75 @@
...
+`<feed feedURL="http://foo.com">
+<xul:label xbl:inherits="value=feedURL,tooltiptext=feedURL">`
+8. Scoped CSS Styling: this can be done with Shadow DOM
+9. Insertion Points: XBL and Shadow DOM appear to use similar models but different keywords. Further investigation is needed.
+10. `[role]`: Used for accessibility: `role="xul:toolbar"`
```
Mossop: Is this XBL specific?

0004-xbl-web-components.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Performance Considerations
+
+XUL and XBL both use fastload and they both have prototype cache, so loading them and opening new windows should be quite fast. Custom Elements (and Shadow DOM) don't have anything like that right now. If we want to use them in browser chrome, we may need to figure out how to support things like this.
+
+The first step is getting some measurements to compare things like element creation time, memory usage, etc. This work is being tracked in [Bug 1387125](https://bugzilla.mozilla.org/show_bug.cgi?id=1387125).
+
+## Notes
+
+* Custom Elements don't currently work in XUL documents, nor are there plans to support them.
```
Mossop: This probably deserves calling out and investigating more. If we plan to remove XBL but not XUL then this could negate the whole plan discussed in this document.

Mossop (changes requested):
I think you should expand on the issue with XUL in its own section before we land this.

0004-xbl-web-components.md
bgrins: Fair point. I gathered this note from a recent discussion with @mrbkap.

Blake, could you elaborate on the reasons why Custom Elements / Shadow DOM wouldn't work in XUL documents? AIUI we'd need some changes to the XML parser, but I believe we'll need to do that for XHTML anyway.

0004-xbl-web-components.md
bgrins: Yes, it can be done with a MutationObserver and a bit of glue. This is prototyped out here: https://github.com/bgrins/xbl-analysis/blob/gh-pages/elements/base-element.js#L29.

I'll add some notes about this in the doc.

0004-xbl-web-components.md
mrbkap: I'm trying to remember more exactly what the collisions are. Off the top of my head, we'd need to deal with XUL prototype documents creating elements not from source (via the XML parser) but from a pre-parsed prototype. When Shadow DOM is implemented, I would be very worried about embedding Shadow DOM inside of XBL anonymous content.

XUL elements are also subtly different from HTML elements in how they handle their parenting and prototype chains.

bgrins: OK, thanks! I'll rephrase this part to something like:

Web Components don't currently work in XUL documents and additional work is required to support them. In particular:
* Elements can be created from the XUL prototype cache, bypassing the XML parser
* XUL elements are subtly different from HTML elements in how they handle their parenting and prototype chains
* Embedding Shadow DOM inside XBL anonymous content is discouraged. If this became necessary it would require more analysis.

Let me know if that needs any updates

0004-xbl-web-components.md
bgrins: The `[role]` attribute isn't, but I don't see the ability to pass a tag name in [the spec](https://www.w3.org/TR/wai-aria/roles#role_definitions). I do see a `toolbar` role though for instance, so I'm inclined to remove this point from the doc.

0004-xbl-web-components.md
```diff
@@ -0,0 +1,75 @@
...
+
+2. Extension Compat: For instance, the Tab Center test pilot addon causes tab browser to become vertical via inheritance with some changes.  This allows us to use the same tag name / markup so that it doesn’t need to re-write CSS. However, with the migration to Web Extensions this won’t be necessary in the general case.
+3. Binding via CSS: `-moz-binding` allows a XBL binding to be attached or detached to any element through CSS.  Web Components would be tied to HTML (based on the particular tag being created). This feature is used in [at least one place to swap an element to a different binding when setting a property](http://searchfox.org/mozilla-central/rev/7cc377ce3f0a569c5c9b362d589705cd4fecc0ac/toolkit/content/widgets/text.xml#47), but we can work around it.
+
+## XBL Usage in Firefox
+
+How many times are individual bindings used?  We've written a module to track how often XBL (and XUL) are used, and a modified version of the test runner that [keeps a count throughout the mochitest-browser suite](https://bugzilla.mozilla.org/show_bug.cgi?id=1376546). Here are initial results:
+
+TODO: ENTER LATEST RESULTS HERE
```
bgrins: I'll just copy the top few here and link to https://bugzilla.mozilla.org/show_bug.cgi?id=1376546 which has the latest results

bgrins:
@Mossop made a round of updates, your review comments should be addressed

Mossop (approved):

PR #7, opened 2017-08-18
Add Architecture Review Process

joewalker:
It's sometimes hard to read a document in a PR. [Here is a rendered version of the Architecture Review Process](https://mozilla.github.io/firefox-browser-architecture/text/0006-architecture-review-process.html).


joewalker:
Previously: https://docs.google.com/document/d/1PtOSBPPewSYtSHmJe8MgJ6yx9T9SYQ2u5-qMqPvdzw8

The only significant change since this document was last reviewed by @dbaron is the addition of the paragraph "_The ideal review process scales well_..." towards the bottom of the Timetable section.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+A Design Review should focus on:
+
+* The solution space
+* Technical roadmaps
+* Milestones and timelines
+* Resource allocation
+
+The key outcome is middle management and individual engineers understanding the value proposition and the path to success.
+
+In [RACI](https://en.wikipedia.org/wiki/Responsibility_assignment_matrix) terms. The team is Responsible, the chair is accountable, other senior engineers or managers are Consulted and various others in the organization are Informed.
```
fluffyemily: nit: Accountable

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+  * Made available to the the team prior to the review so they can prepare answers
+
+4. The chair convenes a review meeting. The discussions of this meeting should be carefully minuted. The meeting comprises:
+  * The team making the proposal
+  * The chair
+  * (For Design Reviews) other senior engineers who will provide valuable input.
+
+The review process should address the question(s) where possible, but the meeting may also decide to alter the the question(s) and undergo a subsequent review. For questions that are go/no-go in nature the decision does not need to be taken during the meeting, although we should be very open about conversations that happen after the review meeting.
+
+The ideal review process scales well. Ideally the same basic system should work for a quick 2 person decision over the best way to design a certain feature, and for a critical organization wide decision about a path to take. The term “team” is used above although we strongly recommend design reviews for smaller questions. If you feel yourself wondering if some design is best, it should be easy to valuable to perform a review.
```
fluffyemily: The continuing repetition of ideal makes it sound as if you don't have confidence in this. Perhaps more positive/definite phrasing?

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+The goal is to have the chair (representing the responsible decision makers) and the team establish some shared reality informally, and then to have the team document this and explain how their work addresses the decision maker’s question within that reality.
+
+The chair then evaluates and refines the team’s document with the assistance of domain experts and the identified stakeholders. Hopefully the consulted stakeholders concur with the framing of the problem and are aligned behind the architectural approach, but perhaps they are not. The chair collects such concerns and presents them. This should be a collaborative process whereby the architectural review functions to document a problem and a proposed solution, in which case the chair can solicit response from the team.
+
+Hopefully, involving the team and the chair informally and subsequently sharing ownership of a document will seed a collaborative process. A particular benefit is avoiding drift between the proposal and the focus of the review: that is, a "successful review" that results in a “go” or “no go” from responsible decision makers in which the review doesn't actually evaluate what the team proposes.
```
fluffyemily: `Hopefully` doesn't engender confidence in the outcome of this step. Make it more definite.

fluffyemily (approved):

text/0005-architecture-reviews.md
joewalker: Thanks. Fixed in 97c6452

text/0005-architecture-reviews.md
joewalker: Fixed in 97c6452

text/0005-architecture-reviews.md
joewalker: Fixed in 97c6452

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
+# Architecture Review Process
+
+## Introduction
+
+Architectural Review decisions within the Firefox organization fall into two categories: "Roadmap" and “Design”.
+
+Product Reviews are not the scope of this process. This process is about architectural change, which should ultimately be in support of product features, but will generally have an effect on multiple product features. Architectural change can be growing new skills or in reducing large scale technical debt.
```
rnewman: Nit: large-scale.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+The function of a **Roadmap Review** is to decide if a thing should be done. The goal is to bring together a packet of data to inform a management decision to provide resources to make the thing happen. A Roadmap Review should happen early in the process so that build time isn’t wasted on a "No" decision, but so that enough information is available to management.
```
rnewman: Worth mentioning another phrasing of this: a Roadmap Review is intended to determine whether there is consensus on (a) whether a problem exists, (a.1) what the problem is, (b) whether it should be solved, (c) if a solution doesn't already exist.

That is: several ideal Roadmap Review outcomes are:

1.a. "No, that is not a problem; it's that way for a reason."
1.b. "That's just a symptom; the real problem is that…"
2. "Yes, that's a problem, but we shouldn't solve it because …"
3. "Yes, that's a problem, and team X already has a solution for a nearly identical problem, so we shouldn't build another".

The last is a good feature of the [Etsy review process](http://www.kitchensoap.com/2017/08/12/multiple-perspectives-on-technical-problems-and-solutions/).

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+A **Design Review** is to decide if the thing should be done in a particular way. The goal is to ensure that the problem space is understood well enough that a realistic bug breakdown can be made, that those working on the problem all agree on the challenge, and that work can be prioritized to hit the hardest parts of the problem first.
```
rnewman: Perhaps "is to decide if a particular proposed approach is the right way to tackle an agreed-upon problem".

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Timetable
+
+For both a Roadmap Review and a Design Review the **timetable** is as follows:
+
+1. Lay the stage. The team should document (See "initial meeting" below):
+  * Someone to **chair** the review (for a Roadmap Review this will be the *least* senior person in the management chain with the ability to provide resources to the problem, for a Design Review this will be the *most* experienced engineer outside of the team willing to put time into the problem. As engineers don’t review their own code they shouldn’t review their own designs)
```
rnewman: Nit: first comma in parenthetical should be a semicolon or full stop.

rnewman: I would suggest breaking out the parenthetical and just making this a paragraph of sentences. You need to address in more detail how a chair is selected — what happens for cross-cutting/cross-team concerns? Do we pick a roadmap chair from just one branch of the org, or does this require Mayo or Beard to chair cross-cutting reviews?

The less senior a RR chair, the less incented they are to declare that something is a problem or to dedicate resources to it, particularly if the outcome would affect other managers' resources and cause tension. Imagine a plain ol' Engineering Manager with quarterly goals and four reports — should they be chairing a review of a proposal to rewrite a large part of their product, just because they have resources to apply to it? They're likely to be biased against a suggestion because of risk aversion.

The more senior a DR chair, the more the availability heuristic applies, particularly if very senior engineers are "willing to put time in" — a Distinguished Engineer might under-weight problems outside of their own domain due to extensive experience with their own problems, or generally downweight the day-to-day costs of front-line engineering from which they've been distanced for several years.

I think the trick is to find a chair — roadmap or design — who has the *appropriate* scope and perspective for the problem or solution being proposed. 

If the problem spans multiple engineering groups, that's probably a senior director and a Principal Engineer. If the problem is industry-scale, that's a VP and Distinguished Engineer. If the problem is confined to a single product or family of products, the chairs might be a director and a Senior Staff Engineer or Principal Engineer. It might be possible to link these criteria to the JFA.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+  * The **question** (or questions) to be resolved, which should be agreed between the team asking for review and the chair. For a Roadmap Review this is likely to be in the form of a Go/No-Go decision. For a Design Review this is likely to be a more open question (e.g. "How should this design be improved")
```
rnewman: As mentioned above, I suspect that shaping RR questions as go/no-go won't be as constructive as it could be, unless the question is a very early stage "is there something to this?".

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+2. The team produces a **Review Packet** designed to answer the question identified in step 1. The production of the Review Packet is expected to be collaborative and iterative to ensure that the review focuses on the key questions and the team presents its best case. This is particularly true of a Roadmap Review where the consequences of the answer are likely to be higher.
```
rnewman: Consequences are more significant, or more costly, etc., not just "higher".

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+The Review Packet includes:
+  * A lay summary of the problem space and the stakeholders which focused on defining a shared language and identifying the key forces behind the problem
```
rnewman: which is

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+  * A brief explaining what the team proposes which should read more like an encyclopedia entry than a marketing document.  The brief should identify:
+    - the forces on the system
+    - the proposed system’s essential architecture
+    - scenarios that exercise the relevant forces against the architecture
+    - how the proposed system handles these scenarios
```
rnewman: Careful with the word "system" here. A RR might propose systemic changes, or the investigation and design of a system, but it probably doesn't propose a system directly.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+  * A competitive analysis suggesting alternatives, costs, and opportunities.
+
+3. The chair creates a list of questions to be discussed during the review. These questions should be:
+  * Open-ended
+  * Made available to the the team prior to the review so they can prepare answers
+
+4. The chair convenes a review meeting. The discussions of this meeting should be carefully minuted. The meeting comprises:
+  * The team making the proposal
+  * The chair
+  * (For Design Reviews) other senior engineers who will provide valuable input.
```
rnewman: I think the goal of not having other engineers in a RR is to avoid ratholing and argument, but I think it's important for a RR to be a discussion of a problem, its prioritization, and also whether other teams in the organization also have that problem (or even already have solutions). With or without other engineers, we need input in order to do that.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+The review meeting should be a discussion of the issues, but should avoid feeling pressured to make a decision, the goal is to understand the issues raised by the question(s) and the background from the review packet, and to add to it wisdom from the people at the review. The meeting may decide to alter the questions and undergo a subsequent review. For questions that are go/no-go in nature the decision does not need to be taken during the meeting, although we should be very open about conversations that happen after the review meeting.
```
rnewman: `s/decision, t/decision. T/`

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+The ideal review process scales well. The same basic system should work for a quick 2 person decision over the best way to design a certain feature, and for a critical organization wide decision about a path to take. The term “team” is used above although we strongly recommend design reviews for smaller questions. If you feel yourself wondering if some design is best, it should be easy to valuable to perform a review.
```
rnewman: I think it's important to re-emphasize here that the goal of this process is to be inclusive, constructive, outcome-seeking, sensitive, empathetic, and open. For both a RR and a DR the goal is to hear perspectives that will lead to a better definition of a problem or solution, and better judgment about whether it should be solved or how to solve it.

Again, I think the Etsy post discusses this in some depth.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+## Rationale
+
+### Key considerations
+
+* What are the desired **outputs** of the review process?
+* Which **audience** is intended to consume the outputs of the review process?
+* What committee **structure** is best positioned to deliver those outputs?
+* What are the **inputs** that the committee should evaluate?
+* What size and **maturity** should projects achieve before undergoing review?
+* When should review results be **disseminated**?
```
rnewman: And:

* What process will most encourage participants to view review as an asset and as an opportunity to seek out, not as an obstacle to avoid?
* What kinds of conduct are necessary in order for the process to be inclusive and constructive, rather than being optimized for certain privileged groups with particular communication styles (_e.g._, verbally argumentative white male senior engineers)?

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+A Design Review should focus on:
+
+* The solution space
+* Technical roadmaps
+* Milestones and timelines
+* Resource allocation
+
+The key outcome is middle management and individual engineers understanding the value proposition and the path to success.
+
+In [RACI](https://en.wikipedia.org/wiki/Responsibility_assignment_matrix) terms. The team is Responsible, the chair is Accountable, other senior engineers or managers are Consulted and various others in the organization are Informed.
```
rnewman: `s/terms. T/terms, t`

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Structure and inputs
+
+The second two considerations shape how the review process actually happens.
+
+The chair and the team frame the question or questions at the start of the process to ensure agreement on what should be tested. The chair is accountable for answering the question or questions following the review process.
+
+In order to agree the question(s), the chair and the team/technical lead or principles will generally have an initial meeting. This meeting should be as small possible and is intended to:
```
rnewman: agree on

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+* establish vocabulary
```
rnewman: Capitalize to match the other bulleted lists.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+
+Involving the team and the chair informally and subsequently sharing ownership of a document is designed to seed a collaborative process. A particular benefit is avoiding drift between the proposal and the focus of the review: that is, a "successful review" that results in a “go” or “no go” from responsible decision makers in which the review doesn't actually evaluate what the team proposes.
+
+### What to review and what to say after review
+
+The final two considerations shape what is reviewed and how the review concludes.
+
+A project under review will be in one of the following states:
+
+* The "Roadmap" stage: the team has an idea, and wants broader buy-in to justify investment in the problem space.
```
rnewman: the team has identified a problem or opportunity

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,135 @@
...
+* The "Design" stage: the problem space is considered valuable, and the team wants to commit to one path through the solution space.
```
rnewman: There's a step missing here.

We go straight from agreeing on the problem to evaluating a single solution. But good projects go from agreeing on the problem to exploring and refining multiple solutions, and actively seeking out existing options, before proposing a single path.

text/0005-architecture-reviews.md
joewalker: I think the Etsy process is much more closely aligned to a Design Review. "The ARWG’s role was _not_ to make a decision" (emphasis theirs), which is the opposite of a Roadmap review.

text/0005-architecture-reviews.md
joewalker: I see the Roadmap/Design distinction as being about:

* resource planning being aimed at management
vs.
* code planning being aimed at engineering.

So it's not a given that we agree that a problem is real before a design review. In fact I hope that the process is lightweight enough that we can use it for any "departures" in Etsy speak.

text/0005-architecture-reviews.md
joewalker: Lets have a 1:1 about this. I read through your examples, and for each, I thought that the proposed algorithm would work. But maybe there's some detail I'm missing

text/0005-architecture-reviews.md
joewalker: Def. supposed to be early stage. The question is "should we invest?".

text/0005-architecture-reviews.md
joewalker: We need more than just a problem, right? Maybe the team needs a 'proposal'.

text/0005-architecture-reviews.md
joewalker: True, but shouldn't we wait for until the team has a proposal before thinking of a review? There is a slot in the review packet for "A competitive analysis suggesting alternatives, costs, and opportunities"

text/0005-architecture-reviews.md
joewalker: Added a paragraph

text/0005-architecture-reviews.md
rnewman: I think it's fair to say it's more holistic (which is both good and bad): it tries (IIRC) to both consider whether there's a problem, and also consider possible outcomes.

The reason I mention here the possibility of having more than a yes/no outcome is that a yes/no outcome treats the RR as an oracle, and I think it's likely that it will naturally end up in a more productive place: as a discussion that's seeded with a coherent set of observations, and ends up with an agreed description of a problem and a possible roadmap.

A draft roadmap might also be _input_, but I see no reason why a negotiated roadmap might not also be _output_.

Do you agree with my (non-exclusive) list of potential outcomes? Am I reading the process correctly?

text/0005-architecture-reviews.md
rnewman: Hm. The reason I make that distinction is that without agreement on the outcome I think many DRs will end up arguing about the problem, not the solution — that is, if you don't think we should be going to the moon, we're going to spend time arguing about that, not about whether my space elevator is better or worse than a shuttle, no? And if there isn't already agreement on the outcome, all of my research into nanowire elevators is wasted.

So I've been reading this document as:

* RR: agree on the problem, at a high level, and a resourcing roadmap to address it.
* DR: given an agreed problem, narrow down a defined outcome and an engineering path to get there.

joewalker:
Worth noting. There's a race for 0005 between this PR and #11. Probably 11 will win.

bgrins (approved):
We've been working through the 'design review' process for XBL removal. I've raised some questions about the process directly in the doc, but I don't see any reason we can't land this now and evolve it as we go. It'd be great to get it landed so we can reference it directly from the gh-pages URL.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+A **Design Review** is to decide if the thing should be done in a particular way. The goal is to ensure that the problem space is understood well enough that a realistic bug breakdown can be made, that those working on the problem all agree on the challenge, and that work can be prioritized to hit the hardest parts of the problem first.
+
+## Timetable
+
+For both a Roadmap Review and a Design Review the **timetable** is as follows:
+
+1. Lay the stage. The team should document (See "initial meeting" below):
+  * Someone to **chair** the review
+    - For a Roadmap Review this will be the *least* senior manager with the ability to address the problem in its entirety.
+    - For a Design Review this will be the an experienced engineer in the problem domain outside of the team willing to put time into the problem. As engineers don’t review their own code they shouldn’t review their own designs.
```
bzbarsky: s/the an/an/

What happens if there is no one outside the team who's experienced in the problem domain?

It seems to me that if someone like that can't be found, then we should try to come up with some process for picking a chair anyway.  I'm not sure I have any concrete suggestions so far...

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+  * The **question** (or questions) to be resolved, which should be agreed between the team asking for review and the chair.
```
bzbarsky: Should probably be a '-' instead of a '*' to make this be part of the right list.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+
+### Process and Conduct
+
+Involving the team and the chair informally and subsequently sharing ownership of a document is designed to seed a collaborative process. A particular benefit is avoiding drift between the proposal and the focus of the review.
+
+An Architecture Review process can be formally mandated. This may work in environments with a large degree of formality and process. This isn't how Mozilla works. For an Architecture Review process to be successful at Mozilla it must be beneficial to the team undergoing review.
+
+The primary informal output of a review should be learning. It is expected that authoritarian go/no-go decisions will fall out additional learning rather than being imposed in the form of a question.
```
bzbarsky: The second sentence of this paragraph doesn't quite make sense.  What does "imposed in the form of a question" mean?

Also, I assume "fall out from additional learning"?

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+
+After the review, the chair and responsible decision makers should provide concrete recommendations and examples for the next teams to undergo the review process. The results should be disseminated as widely as possible, with the package documents and recommendation document archived. Some version of the decision should be publicly linked, so that it can be the decision of record, to which blog posts, mailing list posts, the message saying that the GitHub repository is obsolete, etc., can refer. The goal is to avoid critical technical and resourcing decisions being internally settled and only — later, begrudgingly — being made public on some mailing list. Or, per RACI, the goal is to broaden the set of those Informed.
+
+## Additional Reading
+
+### Multiple Perspectives On Technical Problems and Solutions - John Allspaw - 2017
+
+Describes how Etsy's architecture review process evolved. A solid write-up of the social side of an Architecture Review process in an envronment that is likely closer to Mozilla's than many of the other write-ups.
```
bzbarsky: "environment"

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+The key outcome is senior management investing in the problem space and building consensus around a shared reality.
+
+A Design Review should focus on:
+
+* The solution space
+* Technical roadmaps
+* Milestones and timelines
+* Resource allocation
+
+The key outcome is middle management and individual engineers understanding the value proposition and the path to success.
```
bzbarsky: For a design review, I think a key outcome is engineers understanding the design constraints involved and possibly the tradeoffs between them.  This is especially important if implementation reality leads to having to re-evaluate parts of the design.  Having clear documentation for why the design is the way it is helps avoid duplicating a bunch of work in that situation.

text/0005-architecture-reviews.md
```diff
@@ -0,0 +1,163 @@
...
+The audience of a Roadmap Review is those "up the chain" of decision makers and the audience of a Design Review is those running “down the chain” of implementers.
+
+A Roadmap Review should focus on:
+
+* The problem space
+* Large forces impacting Mozilla’s choices in the space
+* Competitive analysis
+* Opportunity cost
+
+The key outcome is senior management investing in the problem space and building consensus around a shared reality.
```
bzbarsky: I think the other key outcome is providing some indication of how success of the project can be evaluated down the line, as well as documentation of what the fundamental assumptions (e.g. about market structure) the project is based on are.

The latter is especially important if it turns out, as the project progresses, that some of those fundamental assumptions are incorrect, or become incorrect.  Understanding what they are should provide us built-in triggers that might trigger a re-evaluation of the roadmap.

text/0005-architecture-reviews.md
joewalker: Nit: Fixed.

I added "_The principle is to get the most informed and least biased feedback possible_" as a stab at guidelines for selection.

text/0005-architecture-reviews.md
joewalker: Yeah. The intent was that the question was something to be documented. I'll reword to make this clearer.

text/0005-architecture-reviews.md
joewalker: Agreed.

> The key outcome is firstly, senior management investing in the problem space and building consensus around a shared reality. Secondly, to provide documentation of both the goals and assumptions of the project to aid later evaluation.

text/0005-architecture-reviews.md
joewalker: Yes, I think that makes sense. I put:

> The key outcomes are for middle management to understand the value proposition and the path to success and for engineers to understand the design constraints and the tradeoffs between them (this latter part is important when implementation reality causes a re-think, in which case we want to avoid unnecessary re-work).

text/0005-architecture-reviews.md
joewalker: The intent was to say - Rather than asking "_Should we invade Russia in the winter: Yes/No_", and then get into an entrenched, heated debate, we instead ask "_Evaluate the plan to invade Russia in the winter_" and then learn about history and geography to jointly decide that there are better ideas.

I've written:

> The primary informal output of a review should be learning. It is expected that go/no-go decisions will be the result of additional learning rather than the team seeing a 'no-go' as something that needs fighting. Imposed decisions may be required, but learning is a preferred outcome.

Also, nit (s/out/fall out/): Fixed.

joewalker:
This document is incomplete, however we decided to add it to the repo before it's finished because we would like a couple of trial runs before we say we're happy, and we would like something to refer to during those reviews.

It has been marked incomplete, it will either be fixed and completed or removed in subsequent PRs.

text/0006-architecture-review-process.md
```diff
@@ -0,0 +1,172 @@
...
+
+**Warning**: This document is incomplete. We've added it to the repo before it's finished because we would like a couple of trial runs before we say we're happy, and we would like something to refer to during those reviews.
+
+# Architecture Review Process
+
+## Introduction
+
+Architectural Review decisions within the Firefox organization fall into two categories: "Roadmap" and “Design”.
+
+Product Reviews are not within the scope of this process. This process is about architectural change, which should ultimately be in support of product features, but will generally have an effect on multiple product features. Architectural change can be growing new skills or in reducing large-scale technical debt.
```
erikrose, line 16: I'd suggest "skills" → "capabilities", unless you're talking about building people's skills.

text/0006-architecture-review-process.md
```diff
@@ -0,0 +1,172 @@
...
+
+The function of a **Roadmap Review** is to decide if a thing should be done. The goal is to bring together a packet of data to inform a management decision to provide resources to make the thing happen. A Roadmap Review should happen early in the process so that build time isn’t wasted on a "No" decision, but so that enough information is available to management.
+
+A **Design Review** is to decide if the thing should be done in a particular way. The goal is to ensure that the problem space is understood well enough that a realistic bug breakdown can be made, that those working on the problem all agree on the challenge, and that work can be prioritized to hit the hardest parts of the problem first.
```
erikrose, line 20: > hardest parts…first

I can think of 2 motivations for this: to get early warning if the hard parts prove impossible or to inform schedule projections early (assuming hard things take the longest). Are these the intents? Either way, I wouldn't mind having the motivations made explicit; so often, it's nice to have the "spirit of the law" written down to function as a guide in exceptional times.

text/0006-architecture-review-process.md
```diff
@@ -0,0 +1,172 @@
...
+  * A brief that explains what the team is proposing. This should read more like an encyclopedia entry than a marketing document. The audience is the people in the review; i.e. this should attempt to plug organizational documentation holes. The documentation process should not be more onerous than is required for the review. The brief should identify:
+    - the forces on the system
+    - the essential architecture behind the proposal
+    - scenarios that exercise the relevant forces against the proposed architecture
+    - how the proposal handles these scenarios
+  * A competitive analysis suggesting alternatives, costs, and opportunities
+
+3. The chair creates a list of questions to be discussed during the review. These questions should be:
+  * Open-ended
+  * Made available to the the team prior to the review so they can prepare answers
```
erikrose, line 44: Amen. Hooray for agendas ahead of time, which fend off evidence-free exchanges.

text/0006-architecture-review-process.md
```diff
@@ -0,0 +1,172 @@
...
+
+4. The chair convenes a review meeting. The discussions of this meeting should be carefully minuted. The meeting comprises:
+  * The team making the proposal
+  * The chair
+  * Other senior engineers who will provide valuable input
+
+The review meeting should be a discussion of the issues, but should avoid feeling pressured to make a decision. The goal is to understand the issues raised by the question(s) and the background from the review packet, and to add to it wisdom from the people at the review. The meeting may decide to alter the questions and undergo a subsequent review. We should be very open about conversations that happen after the review meeting.
+
+The ideal review process scales well. The same basic system should work for a quick 2 person decision over the best way to design a certain feature, and for a critical organization-wide decision about a path to take. The term “team” is used above although we strongly recommend design reviews for smaller questions. If you feel yourself wondering if some design is best, it should be easy to perform a review.
```
erikrose, line 53: I find lightweight design reviews are well-embodied as (for example) GitHub PRs with no code attached. Then, if we decide to go ahead, we add code. Though this might not work as well in Bugzilla-centric projects since Bugzilla descriptions are immutable, making it hard to maintain a running summary and leading to Death By 1000 Comments.

text/0006-architecture-review-process.md
joewalker: Agreed. Now:

> ... to hit the hardest parts of the problem first (in order to get early warning if the hard parts prove impossible or to inform schedule projections early).

text/0006-architecture-review-process.md
joewalker: Similar - We've found that in the early stages, even a PR is too heavy-weight and we've been using GDocs until we're 80% there. The remaining 80% is done in a mix of PRs, and bugzilla meta bugs.

text/0006-architecture-review-process.md
joewalker: Fixed in b7d27534

text/0006-architecture-review-process.md
joewalker: Fixed in b7d27534

PR #8, opened 2017-08-18
Add a basic Mozilla theme

joewalker:


bgrins:
@joewalker can you give instructions on how to build the repo into a website so I can test this?

joewalker:
Honestly, the way I've been testing is to push to the gh-pages branch in my fork, which gives me http://incompleteness.me/firefox-browser-architecture/

I had a quick hack at jekyll and decided that `git push` was a lot easier. Maybe had I seen https://help.github.com/articles/setting-up-your-github-pages-site-locally-with-jekyll/ earlier or https://jekyllrb.com/docs/github-pages/ and https://jekyllrb.com/docs/quickstart/ then I might have stuck with the command line.

assets/css/style.scss
```diff
@@ -0,0 +1,124 @@
...
+table {
+  margin-bottom: 1.25em;
+}
+
+a {
+  color: #006688;
+}
+
+a:hover, a:focus, a:active, a:visited:hover, a:visited:focus, a:visited:active {
+  color: #0088aa;
```
bgrins, line 114: Can we also do `font-weight: normal` here to prevent size changing when hovering links?

bgrins: (which appears to be coming from site.theme)

joewalker: Fixed in 2849291

_layouts/default.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<html lang="{{ site.lang | default: "en-US" }}">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="chrome=1">
```
bgrins: I don't think we need to use chrome frame :)

_layouts/default.html
```diff
@@ -0,0 +1,18 @@
...
+    <link rel="stylesheet" href="{{ '/assets/css/style.css?v=' | append: site.github.build_revision | relative_url }}">
+    <meta name="viewport" content="width=device-width">
+    <!--[if lt IE 9]>
```
bgrins: ..nor does this page use HTML5 tags

_layouts/default.html
```diff
@@ -0,0 +1,18 @@
...
+    <script src="//html5shiv.googlecode.com/svn/trunk/html5.js"></script>
+    <![endif]-->
+  </head>
+  <body>
+    <div class="wrapper">
+      {{ content }}
+    </div>
+    <script src="{{ '/assets/js/scale.fix.js' | relative_url }}"></script>
```
bgrins, line 12: This script doesn't exist - the reference can be removed

bgrins: Nevermind, I didn't realize how the theme was working - looks like it does indeed exist at https://github.com/pages-themes/minimal/blob/master/assets/js/scale.fix.js

bgrins (approved):
The theming system is a little confusing, and it makes building the site locally difficult, but that's not because of this change. Ultimately I'd prefer to flatten all of the required assets from `jekyll-theme-minimal` and remove the reference from _config.yml but that's no reason to stop progress here.

_layouts/default.html
joewalker: Fixed in 3b668d4

_layouts/default.html
joewalker: Fixed in 3b668d4

PR #9, opened 2017-08-18
Update README.md

ghost:


mykmelez:
Thanks for the pull request! I'm not sure the sections in this document really need this amount of distinction, however. Each one already has a title with a much larger font size and a thin rule underneath, which seems sufficient to distinguish them. And the thick line that is created by an &lt;hr> is so bold that it feels like it distracts my eyes, making it harder to read the text.

So I think we should keep the styling the way it is. However, I appreciate the contribution and hope this doesn't discourage you from contributing in the future!

PR #10, opened 2017-08-21
clarify proposal and conclusion

mykmelez:
Per [my recent comment in a thread on the original pull request](https://github.com/mozilla/firefox-browser-architecture/pull/2#discussion_r134298508), I think we should revisit the titles of the Problem and Proposal sections, and the structure of the text in the Problem section, in this document. Specifically:
* rename Problem section to Proposal
* change "problem statement" from question to assertion
* rename Proposal section to Conclusion

### DIFF
--- a/text/0002-extracting-necko.md
+++ b/text/0002-extracting-necko.md
@@ -1,6 +1,6 @@
-# Problem
+# Proposal
 
-Can we break out a Gecko component, such as Necko, for reuse in background services and on mobile platforms?
+We should extract Necko for reuse in background services and on mobile platforms.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


text/0002-extracting-necko.md
```diff
@@ -1,6 +1,6 @@
-# Problem
+# Proposal
 
-Can we break out a Gecko component, such as Necko, for reuse in background services and on mobile platforms?
+We should extract Necko for reuse in background services and on mobile platforms.
```
rnewman, line 3: Add to line 9 to fill out some of the previous implicit motivation:

> Furthermore, we expect the question of reuse to also arise for other Gecko components.

rnewman (approved):

PR #11, opened 2017-08-22
Adding links to the Firefox Data Stores documentation from Browser Ar…

fluffyemily:
…ch repo


text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,20 @@
...
+
+We therefore need to look at all of the data stores that we are currently using and document their purpose, structure, the kind of data that they hold and how they share that data.
+
+
+# Solution
+
+The extensive documentation can be found in the [Firefox Data Stores](https://github.com/mozilla/firefox-data-store-docs) repository.
+
+* There are a total of 46 separate data stores in Firefox for desktop. Within these 46 stores, there are 10 different storage formats. Very few of these stores are well documented in the code. 
+* Firefox for Android has only 2 stores on the frontend, but shares many of the same stores Gecko stores.
```
joewalker: "_stores Gecko stores_" is a bit clumsy but more importantly it sounds like you're saying it both only has 2 stores, but also many of the 46

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,20 @@
...
+* Firefox for iOS has 5 data stores. One of these is historical as the Reading List store was reused from a another Mozilla app. One of the stores is used to store metadata for use in Activity Stream and has been created separately to make syncing of this data at a later data easier. 
+* Many of the data stores on desktop are synchronous, for example prefs.js. prefs.js is a simple key value file written to disk, but it contains well over 1000 entries. It's synchronous nature means that every time a value is changed every single value is written out. For a data store that is the default for storing single values for the entire app, this proves to be highly inefficient.
```
joewalker: I think the implementation uses a synchronous API, but the implementation is asynchronous. So if Firefox crashes in-between someone setting a pref and the flush happening...
Also, whilst it's true that the whole DB is written for a single pref change, this isn't a result of it's synchronous nature. That's as a result of the poor storage format.

joewalker (commented):
@rnewman should take a look too

joewalker:
There's a huge amount of work in the data storage documentation. Thanks Emily.

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
+
+# Problem
```
rnewman: I don't think this really needs Problem/Solution headers. "Introduction", if you wish.

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
...
+There are a large number of data stores being used to hold data across all Firefox instances and platforms. There is currently nowhere that documents all these data stores and how they share information between them. We have 3 platforms, desktop, Android and iOS and each of these platforms shares data with the others. But not all the information is synced completely across all platforms, and the data itself is not stored in the same way everywhere. When starting to think about how we might provide a holistic user data solution for Mozilla, it is important to understand the scope of the problem and the specifics of the data that is being handled. 
+
+We therefore need to look at all of the data stores that we are currently using and document their purpose, structure, the kind of data that they hold and how they share that data.
+
+
+# Solution
+
+The extensive documentation can be found in the [Firefox Data Stores](https://github.com/mozilla/firefox-data-store-docs) repository.
+
+* There are a total of 46 separate data stores in Firefox for desktop. Within these 46 stores, there are 10 different storage formats. Very few of these stores are well documented in the code.
```
rnewman: *At least* ten different — it's arguable whether some random key-value format in a .txt file should be considered the same as some other textual format, and RDF/XML is not just XML.

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
...
+
+We therefore need to look at all of the data stores that we are currently using and document their purpose, structure, the kind of data that they hold and how they share that data.
+
+
+# Solution
+
+The extensive documentation can be found in the [Firefox Data Stores](https://github.com/mozilla/firefox-data-store-docs) repository.
+
+* There are a total of 46 separate data stores in Firefox for desktop. Within these 46 stores, there are 10 different storage formats. Very few of these stores are well documented in the code. 
+* Firefox for Android has only 3 stores as it consolidates many of the separate stores found in desktop Firefox into a single SQLite database. It uses only 3 different storage formats, SQLite, .json & .js. Android documents it's data stores well and uses a single file to define the schema for all it's tables.
```
rnewman: This isn't really true.

Firefox for Android uses Gecko's form history store, Gecko prefs, and the vast majority of the same files as desktop.

The differences are:

- It uses the SQLite logins store instead of the modern JSON one.
- It uses its own browser DB (and perhaps metadata DB) instead of `places.sqlite` and `favicons.sqlite`.
- It adds several Android XML preference files ("SharedPreferences").
- `times.json` and `signedInUser.json` are written by Java code in addition to, or instead of, Gecko code.
- It encapsulates most of its data handling code in a smaller set of places, typically behind a `ContentProvider` interface.
- It typically doesn't expose as many interfaces for editing Gecko-side data, so the Android codebase typically doesn't mention them.

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
...
+* Firefox for iOS has 5 data stores. One of these is historical as the Reading List store was reused from an earlier Mozilla app. One of the stores is used to store metadata for use in Activity Stream and has been created separately to make syncing of this data at a later data easier. It uses 2 different storage formats, SQLite and .json. iOS stores are not especially well documented, but the schemas are defined in a series of table or database specific files which makes discovering the schemas trivial. The exception to this is the reading list store which is defined separately from the others due to it history as part of another application.
```
rnewman: The Reading List store was built specifically for Firefox, IIRC; it was just built in a different repo for modularity's sake.

The AS metadata DB is dead now, I think.

Don't forget that we store data in `.plist` files _and_ we put credentials in the iOS system keychain.

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
...
+* Desktop stores logins in an encrypted json store, whereas Android has combined the logins table with it's main SQLite database. iOS keeps it as a separate SQLite database that uses sqlcipher for encryption.
```
rnewman: I don't believe the Android part is true.

We are part-way to removing our dependency on the Gecko-side logins store, but we aren't finished yet, and it doesn't (as far as I know!) put logins into `browser.db`.

See the commit message here:

https://hg.mozilla.org/mozilla-central/rev/2c6222ddc033

text/0005-firefox-data-stores.md
```diff
@@ -0,0 +1,21 @@
...
+* Firefox for iOS has 5 data stores. One of these is historical as the Reading List store was reused from an earlier Mozilla app. One of the stores is used to store metadata for use in Activity Stream and has been created separately to make syncing of this data at a later data easier. It uses 2 different storage formats, SQLite and .json. iOS stores are not especially well documented, but the schemas are defined in a series of table or database specific files which makes discovering the schemas trivial. The exception to this is the reading list store which is defined separately from the others due to it history as part of another application.
+* Desktop stores logins in an encrypted json store, whereas Android has combined the logins table with it's main SQLite database. iOS keeps it as a separate SQLite database that uses sqlcipher for encryption. 
+* Prefs.js is a simple key value file written to disk, but it contains well over 1000 entries. It's design means that every time a value is changed every single value is written out. For a data store that is the default for storing single values for the entire app, this proves to be highly inefficient.
```
rnewman: Nit: its.

N.B., usually the file _is not_ written out when a value is changed! And that's **bad** — it means that without a clean shutdown, prefs isn't durable, but pretends that it is.

rnewman (changes requested):

text/0005-firefox-data-stores.md
fluffyemily: Ok, now I'm kinda confused.

We have `logins.json` which is desktop clearly uses & which migrates data over from `signons.sqlite`.

We have the tables `logins`, `logins_disabled_hosts` and `deleted_logins` in `browser.db` on Android and which is defintely being written to by the frontend (`LoginsProvider.java`). 

I am looking at master, and from that link you posted, this stuff is not yet in production, but in that case what should I document here? The current development state, the current deployed state or the desired future state?

fluffyemily: Or perhaps, what I actually need to document, is that this is in flux and make all 3 states explicit here (and in the data stores document).

rnewman: `logins.json` is indeed migrated from `signons.sqlite` for desktop users bringing profiles forward from a pre-2014 version of Firefox, or in cases of bad dataloss (which, given the total lack of integration with Sync, is perhaps the worst thing we could do — it's a silent rollback, and attempts to fix it by futzing with Sync preferences will erase the new data on the server!).

On Android we have _two_ login stores currently in the tree. We have `PasswordsProvider`, which uses SQLite to share the Gecko-side `signins.sqlite`. And we have `LoginsProvider`, which is checked in and tested but not used. As mentioned in the commit message:

> but for now we just want to get the new provider landed and the tests running

I think we should primarily document the stores that are used, not the future migration path or the past migration paths. By all means mention the future, but it's not quite here yet.

rnewman: To double-clarify: `PasswordsProvider` (and its Gecko-side counterpart `storage-mozStorage.js`) are the ones that are currently used on Android.

joewalker:
Worth noting: There's a race for 0005 between this PR and #7.

fluffyemily:
This is turning out to be a useful summary. I will add this to the original documentation once we have agreed on the words.

joewalker:
> Worth noting: There's a race for 0005 between this PR and #7.

Actually we'll need to use 0007 for this.

PR #12, opened 2017-08-22
adding separators to README.me

ghost:
**this PR does basically aim at:**
- *adding separators to Mozilla README.md;*

---

**the motivation for that lies within:**
- *making the file easier to read;*
- *making the file more clear to be read;*
- *beautifying/stylizing the file;*
- *making the file easier to be understood;*


joewalker:
Thanks for the pull request! You've opened basically the same PR before, so the response this time is similar:

> I'm not sure the sections in this document really need this amount of distinction, however. Each one already has a title with a much larger font size and a thin rule underneath, which seems sufficient to distinguish them. And the thick line that is created by an <hr> is so bold that it feels like it distracts my eyes, making it harder to read the text.
>
> So I think we should keep the styling the way it is. However, I appreciate the contribution and hope this doesn't discourage you from contributing in the future!

ghost:
my apologies. i hadn't read the comment. kind regards. sometimes there are also too many mozilla repositories i contribute to, so difficult to keep track of everything. tiago

PR #16, opened 2017-08-24
Fix link to template

emceeaich:
So it works when repo is published as a GitHub site

This is imperfect since it breaks if the repo moves off of GitHub.


text/0001-documenting-output.md
```diff
@@ -20,7 +20,7 @@ Browser Architecture team will document its output like this:
 
 1. The discussion phase involves conversations in the browser-arch [Google Group](https://groups.google.com/a/mozilla.com/forum/#!forum/browser-arch), [IRC channel](https://www.irccloud.com/#!/ircs://irc.mozilla.org:6697/%23browser-arch), [Slack channel](https://mozilla.slack.com/messages/C5F80LV0C/), etc - when a semi-permanent record of current thinking is needed, we use GDocs, stored in [the Browser Architecture folder](https://drive.google.com/drive/u/1/folders/0BzQINYlY78CtbGVqVDVlZlNJX0k).
 2. When we've come to some internal consensus (note this doesn't mean that the proposal is final) we convert our GDocs to Markdown/Git to create an 'RFC' and get feedback from management and the wider Mozilla developer community as appropriate.
-  * Use [this template](../0000-template.md)
+  * Use [this template](https://github.com/mozilla/firefox-browser-architecture/blob/master/text/0000-template.md)
```
joewalker: Does it work if the link is a single `.`?
I think this is my fault for not noticing this link when I moved the template into the `text` folder.

joewalker:
Also - thanks for the pull request!

text/0001-documenting-output.md
emceeaich: Probably the thing to do would be to fix all the links so that they don't use `../`. 

I'll fix this link so it's just relative to folder, because that's the more obvious fix and I plead a lack of ☕️ .

joewalker:
Thanks Emma. Merged.

PR #27, opened 2017-09-08
add glossary to architecture review process

mykmelez:
I found some of this terminology puzzling until I read Grady Booch's article on the subject. Others may be in the same boat, and since the article isn't free, it'd be helpful to define these terms.


joewalker (commented):
Thank-you for doing this. The terminology was clearly confusing.
I wonder if we shouldn't try to use better terminology rather than define what was externally imposed?

Maybe: "the forces on the system" is code for the "functional and non-functional requirements", and "essential architecture behind the proposal" is code for "how the system meets these requirements"?

mykmelez:
> Thank-you for doing this. The terminology was clearly confusing.
> I wonder if we shouldn't try to use better terminology rather than define what was externally imposed?

That's probably a better approach, if we can find the right terminology. I actually quite like Booch's terminology, which is terse and evocative. But it does seem to require having read his article (or perhaps a book on ATAM architecture design), and I don't think we want to require that for participants in this process.

> Maybe: "the forces on the system" is code for the "functional and non-functional requirements", and "essential architecture behind the proposal" is code for "how the system meets these requirements"?

In Booch's description, "forces on the system" excludes functional requirements but includes business goals. And the "essential architecture" is more the architecture itself rather than how it meets the requirements (the latter of which appears to be the outcome of exercising the forces against the scenarios).

In e230548 I've removed the Glossary and replaced "forces on the system" and "essential architecture" with more descriptive phrases (plus examples). I'm unsure about the "business goals" item, which Booch doesn't really explain, but which I imagine to mean the kinds of outcomes that a product manager is looking to achieve.

How does this branch look now?

joewalker (commented):
I suppose the question surrounding "Business goals" is the level they're talking about. I'd assumed that it was about the "business goals for the system under review" (hence the equality with functional requirements). I think your assumption is that it's about "the goals for Mozilla's business".

I think that both are relevant. I see this broadly as "document the shape of the hole you're filling and match it up to the shape of the peg you're designing". The shape of the hole is defined by both types of business goal.

Maybe we can assume that the team has considered functional requirements, and so the review needs to focus on the wider context that they might have missed.

mykmelez:
> I suppose the question surrounding "Business goals" is the level they're talking about. I'd assumed that it was about the "business goals for the system under review" (hence the equality with functional requirements). I think your assumption is that it's about "the goals for Mozilla's business".

You're exactly right, that's what I thought Booch meant by "business goals". I hadn't even considered that he might be using the term to mean functional requirements!

I went back and re-read the article, but I didn't gain any more insight, except that I noticed a use of the term "business drivers" that appears to be a synonym.

So I did some googling and found this description of the [Architecture Tradeoff Analysis Method](http://www.sei.cmu.edu/architecture/tools/evaluate/atam.cfm) (ATAM), which references the report [Categorizing Business Goals for Software Architectures](http://resources.sei.cmu.edu/library/asset-view.cfm?assetID=7751), which is available as a free PDF and describes business goals thusly:

> Business goals are the foundation on which software systems are justified, analyzed, and built. Software systems are constructed to realize business or mission goals. Software architecture is the bridge between the business goals and the realized system.

It then organizes business goals into five broad categories:

1. reduce total cost of ownership
2. improve capability/quality of system
3. improve market position
4. support improved business processes
5. improve confidence in and perception of the system

And it includes a visual—Figure 3: The Extended Architecture Business Cycle—that differentiates between Quality Attribute Requirements, Business Requirements, and Functional Requirements.

So it seems like business goals are more like the "goals for Mozilla's business" than functional requirements, at least in ATAM.

Note: I haven't read the whole report yet, I've only skimmed it, so I might be missing something!

> Maybe we can assume that the team has considered functional requirements, and so the review needs to focus on the wider context that they might have missed.

I think that's right. I can imagine exceptions, but it seems like an architecture's fitness for its functional requirements should be determined at an earlier stage of design (f.e. via analysis and prototyping by the tech lead for the project).

Perhaps that's worth digging into deeper, however, via more research and a better understanding of ATAM and related methods.

joewalker (approved):
r+ on the grounds that even if *business goals* are only externally defined, this is still an improvement on what was there before.

PR #31, opened 2017-09-19
Fix links in newsletter post.

Mossop:
I think the bare tags here were breaking the markdown parser, this might fix it.


bgrins (approved):
Thanks

PR #37, opened 2017-09-22
Add an RSS feed for newsletter posts

Mossop:
The other newsletters are published on planet.mozilla.org. It might make sense for us to do the same and we can with a bit of tinkering with the repo structure. What do you think of this @joewalker?


newsletters.xml
```diff
@@ -0,0 +1,20 @@
+---
+layout: null
+---
+<?xml version="1.0" encoding="UTF-8" ?>
+<rss version="2.0">
+<channel>
+  <title>Browser Architecture</title>
```
ncalexan: Should this feed say more (Mozilla? Firefox?) for consumers _outside_ of `planet.mozilla.org`?

Mossop: Good point. Probably yes

Mossop: Addressed in 23d494e38fb269584d6eab2248328d8e4a3cf5b3

joewalker (approved):

PR #40, opened 2017-10-24
Updates to arch review doc following XBL removal review

joewalker:
@bgrins and @mykmelez because they were there and @rnewman because he's doing the next review.

### DIFF
--- a/text/0006-architecture-review-process.md
+++ b/text/0006-architecture-review-process.md
@@ -47,10 +47,15 @@ The Review Packet includes:
   * Open-ended
   * Made available to the the team prior to the review so they can prepare answers
 
-4. The chair or reviewer convenes a review meeting. The discussions of this meeting should be carefully minuted. The meeting comprises:
-  * The team making the proposal
-  * The reviewer (and chair if used)
-  * Other senior engineers who will provide valuable input
+4. The chair or reviewer convenes a review meeting. The discussions of this meeting should be carefully minuted.
+  * The meeting comprises:
+    * The team making the proposal
+    * The reviewer (and chair if used)
+    * Other senior engineers who will provide valuable input
+  * A suggested agenda for the review meeting:
+    * Short opening statement by the team on the current state of play
+    * Discussion of Questions and Answers (see 3. above)
+    * A step back to discuss if the proposal is right in the wider context
 
 The review meeting should be a discussion of the issues, but should avoid feeling pressured to make a decision. The goal is to understand the issues raised by the question(s) and the background from the review packet, and to add to it wisdom from the people at the review. The meeting may decide to alter the questions and undergo a subsequent review. We should be very open about conversations that happen after the review meeting.
 
--- a/text/0006-architecture-review-process.md
+++ b/text/0006-architecture-review-process.md
@@ -36,9 +36,9 @@ The Review Packet includes:
   * A list of the groups directly affected by the proposal to ensure that the review meeting includes representatives from those groups.
   * A vision of what the project will achieve on completion
   * A brief that explains what the team is proposing. This should read more like an encyclopedia entry than a marketing document. The audience is the people in the review; i.e. this should attempt to plug organizational documentation holes. The documentation process should not be more onerous than is required for the review. The brief should identify:
-    - the non-functional requirements, like performance, security, and stability
-    - the business goals, such as improving user satisfaction and retention
     - the architecture being proposed
+    - the non-functional requirements, like accessibility, performance, security, and stability
+    - relevant business goals, such as improving user satisfaction and retention
     - scenarios that exercise the requirements/goals against the proposed architecture
     - how the proposal handles these scenarios
     - what review and discussion of the proposed architecture has already happened (and with whom)
--- a/text/0006-architecture-review-process.md
+++ b/text/0006-architecture-review-process.md
@@ -33,6 +33,7 @@ For both a Roadmap Review and a Design Review the **timetable** is as follows:
 2. The team produces a **Review Packet** designed to document the proposal identified in step 1. The production of the Review Packet is expected to be collaborative and iterative to ensure that the review focuses on the key issues so the team presents its best case. This is particularly true of a Roadmap Review where the consequences of the review are likely to be more significant.
 The Review Packet includes:
   * A lay summary of the problem space which is focused on defining a shared language and identifying the key forces behind the problem
+  * A vision of what the project will achieve on completion
   * A brief that explains what the team is proposing. This should read more like an encyclopedia entry than a marketing document. The audience is the people in the review; i.e. this should attempt to plug organizational documentation holes. The documentation process should not be more onerous than is required for the review. The brief should identify:
     - the non-functional requirements, like performance, security, and stability
     - the business goals, such as improving user satisfaction and retention
--- a/text/0006-architecture-review-process.md
+++ b/text/0006-architecture-review-process.md
@@ -33,6 +33,7 @@ For both a Roadmap Review and a Design Review the **timetable** is as follows:
 2. The team produces a **Review Packet** designed to document the proposal identified in step 1. The production of the Review Packet is expected to be collaborative and iterative to ensure that the review focuses on the key issues so the team presents its best case. This is particularly true of a Roadmap Review where the consequences of the review are likely to be more significant.
 The Review Packet includes:
   * A lay summary of the problem space which is focused on defining a shared language and identifying the key forces behind the problem
+  * A list of the people directly affected by the proposal to ensure that the review meeting (see below) includes valuable representatives.
   * A vision of what the project will achieve on completion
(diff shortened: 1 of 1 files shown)


bgrins:
There are two points in the brief that weren't clear to me, at least for the XBL design review:

* scenarios that exercise the requirements/goals against the proposed architecture
* how the proposal handles these scenarios

I'm still not exactly sure what those mean, and we ended up leaving that blank in the packet. Also FWIW we folded these into a single header: "Scenarios that exercise relevant forces against the architecture, and how the proposal handles them"

text/0006-architecture-review-process.md
```diff
@@ -11,27 +11,29 @@ layout: text
...
     - For a Design Review this will be an experienced engineer in the problem domain outside of the team willing to put time into the problem. As engineers don’t review their own code they shouldn’t review their own designs. The principle is to get the most informed and least biased feedback possible.
+  * Optionally, for larger reviews, find someone to **chair** the review. This enables to reviewer to concentrate on the problem rather than the admin of the review.
 
 2. The team produces a **Review Packet** designed to document the proposal identified in step 1. The production of the Review Packet is expected to be collaborative and iterative to ensure that the review focuses on the key issues so the team presents its best case. This is particularly true of a Roadmap Review where the consequences of the review are likely to be more significant.
 The Review Packet includes:
-  * A lay summary of the problem space and the stakeholders which is focused on defining a shared language and identifying the key forces behind the problem
+  * A lay summary of the problem space which is focused on defining a shared language and identifying the key forces behind the problem
```
bgrins, line 35: We did include a 'stakeholders' section for the XBL review, and I think having that was valuable. But maybe it should be pulled out into a separate bullet

bgrins (approved):
Seems like reasonable changes to me. Just a couple of comments above

text/0006-architecture-review-process.md
```diff
@@ -11,27 +11,29 @@ layout: text
...
 
 ## Timetable
 
 For both a Roadmap Review and a Design Review the **timetable** is as follows:
 
-1. The team should set the stage (see "initial meeting" below):
-  * Summarize in a one or two sentences the proposal to be reviewed. This should be agreed between the team asking for review and the chair.
-  * Find someone to **chair** the review:
+1. The team should set the stage:
+  * Summarize in a one or two sentences the proposal to be reviewed. This should be agreed between the team asking for review and the reviewer.
```
mykmelez: While you're here, nit: can you change "in a one or two sentences" to "in one or two sentences"?

text/0006-architecture-review-process.md
```diff
@@ -11,27 +11,29 @@ layout: text
...
+  * Find a **reviewer**:
     - For a Roadmap Review this will be the least senior manager with the ability to address the problem in its entirety.
     - For a Design Review this will be an experienced engineer in the problem domain outside of the team willing to put time into the problem. As engineers don’t review their own code they shouldn’t review their own designs. The principle is to get the most informed and least biased feedback possible.
+  * Optionally, for larger reviews, find someone to **chair** the review. This enables to reviewer to concentrate on the problem rather than the admin of the review.
```
mykmelez: Nit: the phrase "admin of the review" is a bit confusing. If I understand correctly what you're saying, changing it to "administration of the review" would be sufficient.

It might also make sense to clarify that the role of the chair, as distinct from that of the reviewer, is simply to facilitate the process (i.e. schedule the meeting, manage the clock, ensure minutes are taken, etc.).

I also wonder if it's worth elaborating on the difference between a smaller and a larger review. The separation of the roles of review and chair is one difference, but presumably there are others. Ideally, a "small review" is something that can be thrown together quickly and with barely more than a single person "team," a proposal, and a single reviewer.

text/0006-architecture-review-process.md
```diff
@@ -41,14 +43,20 @@ The Review Packet includes:
     - what review and discussion of the proposed architecture has already happened (and with whom)
   * A competitive analysis suggesting alternatives, costs, and opportunities
 
-3. The chair creates a list of questions to be discussed during the review. These questions should be:
+3. The chair or reviewer creates a list of questions to be discussed during the review. These questions should be:
```
mykmelez: Isn't this something that the reviewer should create? After all, it's the reviewer whose questions will need to be addressed. A chair might be able to collate questions from a review panel, but they presumably aren't the subject-matter expert(s) who signed up to generate those questions.

text/0006-architecture-review-process.md
```diff
@@ -41,14 +43,20 @@ The Review Packet includes:
...
   * Open-ended
   * Made available to the the team prior to the review so they can prepare answers
 
-4. The chair convenes a review meeting. The discussions of this meeting should be carefully minuted. The meeting comprises:
-  * The team making the proposal
-  * The chair
-  * Other senior engineers who will provide valuable input
+4. The chair or reviewer convenes a review meeting. The discussions of this meeting should be carefully minuted.
+  * The meeting should have a limited attendance to avoid a sprawling unfocussed meeting, but it should include:
+    * The team making the proposal, the reviewer (and chair if used)
```
mykmelez: Nit: rather than using "chair or reviewer" and caveating references to the chair with "if used", it'd be better to simply define the chair as a permanent role that might happen to be filled by the reviewer. Then you can simply use "chair" when you mean the chair specifically, and that'll apply even in cases where the role of chair is being filled by the reviewer.

mykmelez (approved):
Most notes are nits, except for my comment about who gets to list questions. Still r+ from me, this is a set of welcome improvements.

text/0006-architecture-review-process.md
joewalker: OK. The logic for me removing that was partly the clumsy language, but also that the stakeholders  should know who they are because they are at the review meeting. However on reflection, documenting this does make the thought process more obvious.

I'll add as a separate bullet that the review packet includes "_A list of the groups directly affected by the proposal to ensure that the review meeting includes representatives from those groups._"

See 3d318b3e1850c20794d174fb727180ec42c7aad0 and d4798e3300335c5b4283f884b25c6b5d0072e1c2.

joewalker:
Brian wrote:

> There are two points in the brief that weren't clear to me, at least for the XBL design review:
>
> * scenarios that exercise the requirements/goals against the proposed architecture
> * how the proposal handles these scenarios
>
> I'm still not exactly sure what those mean, and we ended up leaving that blank in the packet. Also FWIW we folded these into a single header: "Scenarios that exercise relevant forces against the architecture, and how the proposal handles them"

The requirements/goals were defined at the top of the list of bullet points but I've changed the order to make this more obvious, and I think we did it both in the packet and the review -- Take performance as the best example, we considered that as an NFR and talked about our impact.

I've made some minor tweaks the language of the bullet points to see if I can make this clearer.

text/0006-architecture-review-process.md
joewalker: See 36ceb71af34aafdda70b939e1d2c58eca0f79dea

text/0006-architecture-review-process.md
joewalker: Clarified wording in 36ceb71af34aafdda70b939e1d2c58eca0f79dea.

The point about differences between a smaller and a larger reviews is made about line 64:

> The ideal review process scales well. The same basic system should work for a quick 2 person decision over the best way to design a certain feature, and for a critical organization-wide decision about a path to take. The term “team” is used above although we strongly recommend design reviews for smaller questions. If you feel yourself wondering if some design is best, it should be easy to perform a review.

Other than the separation of chair/reviewer, most of the changes that I'm aware of would be simply scale, but maybe we need more experience to be sure.

text/0006-architecture-review-process.md
```diff
@@ -11,44 +11,53 @@ layout: text
...
-  * Find someone to **chair** the review:
+1. The team should set the stage:
+  * Summarize in one or two sentences the proposal to be reviewed. This should be agreed between the team asking for review and the reviewer.
+  * Find a **reviewer**:
     - For a Roadmap Review this will be the least senior manager with the ability to address the problem in its entirety.
     - For a Design Review this will be an experienced engineer in the problem domain outside of the team willing to put time into the problem. As engineers don’t review their own code they shouldn’t review their own designs. The principle is to get the most informed and least biased feedback possible.
+  * Optionally, for larger reviews, find someone to **chair** the review. This can facilitate the process (i.e. scheduling the meeting, managing the clock, ensuring minutes are taken, etc.) and enables to reviewer to concentrate on the review itself. The rest of this document uses 'chair' for the administrative role. For smaller reviews the reviewer also does the tasks of the chair.
```
mykmelez: Nit: to reviewer to concentrate -> the reviewer to concentrate.

rnewman (approved):

text/0006-architecture-review-process.md
joewalker: Fixed in 711ee29328f7a86acd4ce746a7a6efb40c416f6d

PR #41, opened 2017-11-07
Doc updates and minor code sample changes

grigoryk:
Apologies in advance for my trigger-happy editors (seems like it's all of them) which declared war on dangling white-space.

As I was going through this post:
- a couple of things confused me a bit
  - see a couple of minor edits to paragraphs and making the `<project_path>` to be consistent in the post
- Rust code didn't actually compile
  - it didn't understand what `rust_greeting_free(world);` was, and that seemed like ungoogleable black magic to me as well
  - According to https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw, my changes should work correctly (and they appear to)
- and once it did compile, it produced a library in `rlib` format, whereas we need an `so`
  - figured out that I can be explicit about the library type!
  - I suspect that this is because I'm running through this on Linux

### DIFF
--- a/experiments/2017-09-21-rust-on-android.md
+++ b/experiments/2017-09-21-rust-on-android.md
@@ -206,9 +206,9 @@ mkdir jniLibs/arm64
 mkdir jniLibs/armeabi
 mkdir jniLibs/x86
 
-ln -s <path>/cargo/target/aarch64-linux-android/release/libgreetings.so jniLibs/arm64/libgreetings.so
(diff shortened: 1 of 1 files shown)


grigoryk:
@fluffyemily mind taking a look?

grigoryk:
Ah, it seems that you forgot to include rust_greeting_free method! (I've read the example code in https://github.com/fluffyemily/cross-platform-rust/blob/master/basic/cargo/src/lib.rs)

My change still feels a bit more obvious (to a person who's seeing Rust for the first time).

fluffyemily (approved):

grigoryk:
@fluffyemily can you press the "merge" button?

PR #46, opened 2017-11-21
fix indentation of list hierarchies

mykmelez:
Apparently child list items must be indented at least as much as their parent item's *text*, which is at least three spaces for numbered parent items, as those are prefixed with three characters: a number (f.e. "1"), a period ("."), and a space (" ").

Thus two-space indentation of child items:

```
1. Foo
  * bar
  * baz
```

Fails to indent the children:

1. Foo
  * bar
  * baz

While three-space indentation:

```
1. Foo
   * bar
   * baz
```

Successfully indents the children:

1. Foo
   * bar
   * baz

Three spaces is arguably uglier than an even amount of indentation when editing the raw text (especially if your text editor assumes even indentation), so this change indents the children of numbered items by *four* characters and the children of bulleted items by *two* characters.

It also makes the bulleting characters more consistent, using an asterisk (*) for a first-level bulleted item, a dash (-) for a second-level item, and a plus sign (+) for a third-level item. (This doesn't affect appearance, as all three bulleting characters result in the same bullets, which vary only by level.)

This PR makes no textual changes. It is strictly formatting.

### DIFF
--- a/text/0006-architecture-review-process.md
+++ b/text/0006-architecture-review-process.md
@@ -24,40 +24,40 @@ A **Design Review** is to decide if the thing should be done in a particular way
 For both a Roadmap Review and a Design Review the **timetable** is as follows:
 
 1. The team should set the stage:
-  * Summarize in one or two sentences the proposal to be reviewed. This should be agreed between the team asking for review and the reviewer.
-  * Find a **reviewer**:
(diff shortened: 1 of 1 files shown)


joewalker (approved):
TIL.
Thanks Myk.

PR #50, opened 2017-12-19
Lightweight design review of Fluent in Preferences

mykmelez:
@rnewman You attended this meeting, so your review of this description of it would be very helpful!

@zbraniecki Your review as an attendee from the engineering team would also be very helpful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ This is a list of our findings that we're reasonably happy with so far.
 * [Roadmap Review: Sync and Storage](text/0008-sync-and-storage-review-packet.md) establishes that storage and syncing of user data is a pillar of the Firefox ecosystem, warranting holistic and long-term attention, and outlines where we’d like to end up and some ideas for how to get there.
 * [JavaScript Type Safety Systems](text/0009-type-safety-systems.md) are some conclusions of an investigation into the use of JavaScript type safety systems.
 * [Firefox Data Stores Documentation](text/0010-firefox-data-stores.md) documents the existing data stores across all current Firefox platforms.
+* [Fluent in Prefs Design Review](text/0011-fluent-in-prefs-design-review.md) describes the lightweight design review for Fluent in Prefs.
 
 ## Posts
 
--- a/text/0011-fluent-in-prefs-design-review.md
+++ b/text/0011-fluent-in-prefs-design-review.md
@@ -1,0 +1,61 @@
+# Fluent in Prefs Design Review
+
+## Introduction
+
+[Fluent](http://projectfluent.org/) is a "family of localization libraries designed to improve how software is translated." Mozilla's localization engineering team (the "engineering team") is migrating Firefox's localization architecture from DTD/properties files to FTL files processed by Fluent, which will enable richer and more accurate translations, runtime locale loading/switching, runtime LTR/RTL and pseudo-locale switching, and a fallback locale. The first target for this migration is the preferences interface (the about:preferences page and its various subdialogs).
+
+The engineering team met with a set of architects and domain experts (the "review team") to discuss risks and mitigations, which constituted a lightweight "design review" of the project's architecture. This document describes the concerns that the review team raised and the risk mitigations that the teams discussed.
+
+## Why Preferences
+
+The team chose preferences because they're text-heavy, include users' names (which exercises Fluent's support for grammatical gender), and because the precision of text matters in preferences to ensure users clearly understand the implications of changes. So preferences are a good way to prove the change to Fluent before adopting it more widely.
+
+Fluent is a big change to the localization process, and the team plans to validate the change by translating just four strings initially (the "name" strings in brand.dtd/properties?) and then verifying those changes across the localization toolchain.
+
+## Chrome Privileges
+
+The review team asked whether there are parts of preferences that don't have chrome privileges, and if that influences the use of Fluent. It seems like the only such part is the Firefox Accounts image (???), which loads over a network and is localized independently of Firefox, so it won't be an issue.
+
+## Search Incompatibility
+
+The search feature of about:preferences is an issue, as the way that it searches preference labels and descriptions is not compatible with the way that Fluent injects text into the page.  The teams discussed options and concluded that there is a straightforward way to update the search implementation to be compatible with Fluent.
+
+Specifically, FTL messages support arbitrary attributes, so an FTL message for a searchable preference control could include a "searchterms" attribute containing the text to search, i.e.
+
+```
+key1
+  .label = "click"
+  .searchterms = " … "
+```
+
+Attribute values can include references to other strings within the FTL file, so localizers could avoid string duplication if desired.  This model would also enable localizers to customize search terms to include extra keywords that aren't included in other strings related to the control.  Finally, this model supports English fallback for searches via the creation of a secondary *document.l10n* object with the `en` locale.
+
+The review team raised some concerns about the performance of runtime searches and suggested that the engineering team consider generating an index from search terms to the controls they match, either at build time or on first search.  Some participants suggested being careful to avoid prematurely optimizing, however. In the end, the teams didn't reach consensus on the optimal solution, and further research/experimentation is warranted here.
+
+## Schedule Risk
+
+After validating the process with the first four strings, the engineering team plans to migrate the rest of the strings, but they're aware of the large set of pending changes for XBL replacement in [bug 1379338](https://bugzilla.mozilla.org/show_bug.cgi?id=1379338), and they'd prefer to await those before proceeding to minimize merge hell.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -26,6 +26,7 @@ This is a list of our findings that we're reasonably happy with so far.
 * [Roadmap Review: Sync and Storage](text/0008-sync-and-storage-review-packet.md) establishes that storage and syncing of user data is a pillar of the Firefox ecosystem, warranting holistic and long-term attention, and outlines where we’d like to end up and some ideas for how to get there.
 * [JavaScript Type Safety Systems](text/0009-type-safety-systems.md) are some conclusions of an investigation into the use of JavaScript type safety systems.
 * [Firefox Data Stores Documentation](text/0010-firefox-data-stores.md) documents the existing data stores across all current Firefox platforms.
+* [Fluent in Prefs Design Review](text/0011-fluent-in-prefs-design-review.md) describes the lightweight design review for Fluent in Prefs.
```
rnewman, line 29: Can we use "Preferences" throughout? There's a slight bias in Firefox to think of `libpref` and associated stuff when we say "prefs", so if we can tweak wording to avoid that confusion, I think it's worth doing.

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+The engineering team met with a set of architects and domain experts (the "review team") to discuss risks and mitigations, which constituted a lightweight "design review" of the project's architecture. This document describes the concerns that the review team raised and the risk mitigations that the teams discussed.
+
+## Why Preferences
+
+The team chose preferences because they're text-heavy, include users' names (which exercises Fluent's support for grammatical gender), and because the precision of text matters in preferences to ensure users clearly understand the implications of changes. So preferences are a good way to prove the change to Fluent before adopting it more widely.
```
rnewman, line 11: Plural agreement: I think "Preferences" the module/interface/feature, short for "the Preferences UI", is singular.

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+Fluent is a big change to the localization process, and the team plans to validate the change by translating just four strings initially (the "name" strings in brand.dtd/properties?) and then verifying those changes across the localization toolchain.
```
rnewman, line 13: Can we Markdown backtick file paths and URIs throughout for clarity?

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Chrome Privileges
+
+The review team asked whether there are parts of preferences that don't have chrome privileges, and if that influences the use of Fluent. It seems like the only such part is the Firefox Accounts image (???), which loads over a network and is localized independently of Firefox, so it won't be an issue.
```
rnewman, line 17: s/image/avatar
s/is localized independently…/is not a localized resource

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+key1
+  .label = "click"
+  .searchterms = " … "
+```
+
+Attribute values can include references to other strings within the FTL file, so localizers could avoid string duplication if desired.  This model would also enable localizers to customize search terms to include extra keywords that aren't included in other strings related to the control.  Finally, this model supports English fallback for searches via the creation of a secondary *document.l10n* object with the `en` locale.
+
+The review team raised some concerns about the performance of runtime searches and suggested that the engineering team consider generating an index from search terms to the controls they match, either at build time or on first search.  Some participants suggested being careful to avoid prematurely optimizing, however. In the end, the teams didn't reach consensus on the optimal solution, and further research/experimentation is warranted here.
```
rnewman, line 33: Worth mentioning that we mostly just agreed that all problems here are solvable, so we didn't feel the need to come to consensus.

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+Coalescing is considered harmful, and a concern was raised that merging preferences strings into a single FTL file might encourage engineers to do it even more. The teams concluded that this concern can be addressed at the policy/review level.
+
+## String Changes
+
+The review team probed about Fluent support for string changes, and the engineering team explained that support is as good as the current system, which has two ways to modify a string, and work is planned to make it even better by adding a third way.
+
+Specifically, under both the current system and the current version of Fluent, a developer can make a minor change to a string (f.e. a typo fix) that shouldn't affect its translations by preserving the string's key, and they can make a major change to a string that should affect its translations by changing the string's key.
+
+In the future, Fluent will support versioning of a string key via a tag, such that developers can also make changes that affect localizations without having to change the string's key. (There was some discussion about which kinds of changes would fit into this bucket.)
```
rnewman, line 61: I believe the Fluent mechanism for this is a special comment, but I don't remember the term; perhaps worth putting it here so we remember in the future.

rnewman (approved):
I seem to recall some brief discussion of branding, which of course applies to Preferences. Could you get a line or two from Zibi to go in this doc?

Thanks for spending the time to write this up!

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
+# Fluent in Prefs Design Review
+
+## Introduction
+
+[Fluent](http://projectfluent.org/) is a "family of localization libraries designed to improve how software is translated." Mozilla's localization engineering team (the "engineering team") is migrating Firefox's localization architecture from DTD/properties files to FTL files processed by Fluent, which will enable richer and more accurate translations, runtime locale loading/switching, runtime LTR/RTL and pseudo-locale switching, and a fallback locale. The first target for this migration is the preferences interface (the about:preferences page and its various subdialogs).
```
zbraniecki, line 5: Fluent is a "localization system designed to..." - http://projectfluent.org/

text/0011-fluent-in-prefs-design-review.md
zbraniecki, line 13: Rough suggestion: "Fluent is a big change to the localization process, and the team is rolling out the change in stages. The first stage landed in Firefox 58 and enabled a single string localized using Fluent (https://bugzilla.mozilla.org/show_bug.cgi?id=1402069). Second stage is about to land (https://bugzilla.mozilla.org/show_bug.cgi?id=1411012) and it migrates four strings to FTL. Each of those is meant to test the complete localization toolchain readiness for the new system. After them, the Preferences migration will begin."

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Schedule Risk
+
+After validating the process with the first four strings, the engineering team plans to migrate the rest of the strings, but they're aware of the large set of pending changes for XBL replacement in [bug 1379338](https://bugzilla.mozilla.org/show_bug.cgi?id=1379338), and they'd prefer to await those before proceeding to minimize merge hell.
+
+The XBL replacement patch author ([myk](https://mozillians.org/en-US/u/myk/)) and reviewer ([jaws](https://mozillians.org/en-US/u/jaws/)) were both members of the review team, and they noted that those changes should land soon and thus shouldn't create schedule risk.
+
+Another member of the review team noted that another team is rewriting the "delete cookies" dialogs, and the engineering team should coordinate with that team.
```
zbraniecki, line 41: can we link to a bug?

text/0011-fluent-in-prefs-design-review.md
```diff
@@ -0,0 +1,61 @@
...
+
+Specifically, FTL messages support arbitrary attributes, so an FTL message for a searchable preference control could include a "searchterms" attribute containing the text to search, i.e.
+
+```
+key1
+  .label = "click"
+  .searchterms = " … "
+```
+
+Attribute values can include references to other strings within the FTL file, so localizers could avoid string duplication if desired.  This model would also enable localizers to customize search terms to include extra keywords that aren't included in other strings related to the control.  Finally, this model supports English fallback for searches via the creation of a secondary *document.l10n* object with the `en` locale.
```
zbraniecki, line 31: `(...) Finally, this model supports adding an English fallback for searches via the creation of a secondary *Localization* object with the same resources registered for the "en" locale.`

text/0011-fluent-in-prefs-design-review.md
zbraniecki: We call it `semantic comments`. The idea is that you could do:

```
key = Value
```

and then later:

```
# @version: 2
key = New Value
```

Runtime would just pick the `key` and ignore the comments, while tooling could notify the localizer that his localization was implicitly marked as `version: 1` and the en-US has `version: 2` so he may want to update the translation.

This is just a concept at this point and I hope to write down the plan for this in Q1.

zbraniecki (approved):

PR #56, opened 2018-02-23
A brief analysis of JSON file-backed storage.

rnewman:
How does this look, @bgrins?


text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+
+[JSONFile.jsm](https://dxr.mozilla.org/mozilla-central/source/toolkit/modules/JSONFile.jsm) is a centralized implementation of this pattern. It supports Promise-based asynchronous loading, synchronous loading, and flushes asynchronously, defaulting to 1500ms after each change. libpref flushes after 500msec.
+
+This approach is well suited to data that:
+
+- Changes infrequently.
+- Is relatively small.
+- Is only ever directly read or written by a single thread.
+- Needs to be kept entirely in memory for reads.
+- Has a straightforward JSON encoding (consider the expense of base64 + UTF-8 for images).
```
ncalexan: nit: I'd elaborate "consider the expense" to state your conclusion.  Perhaps: "For example, base64 + UTF-8 encoding images is an expensive JSON encoding that is avoided in practice."

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- Development is 'natural' for front-end developers — get things working with only in-memory data, then add persistence on top.
+
+Disadvantages:
+- Versioning is often an afterthought: engineers rarely think to version in-memory data formats.
+- Change tracking is almost always an afterthought, thanks to the 'natural' development process: adding syncing later becomes difficult.
+- Users can edit the file on disk, which can result in persistent state that was never created by the component's own logic.
+- Frequent writes will cause the entire file to be written to disk repeatedly, [which causes complaints about SSD impact](https://bugzilla.mozilla.org/show_bug.cgi?id=1304389).
+- The entire file typically needs to be read into memory to be used, which increases memory footprint and can impact startup time.
+- Writes at shutdown harm the user experience and don't happen during a crash.
+- During persistence, the entire file exists twice on disk, which could be problematic on mobile platforms for large data sets.
```
ncalexan: This is an artifact of the approach that is currently taken to achieve atomic writes.  Perhaps say, "In the current implementation" or "The usual implementation" or explain more, in the description of `JSONFile.jsm` itself, why this is the case.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- The only approach to cross-process use is duplication of all or part of the in-memory data, which can increase memory footprint. It's not possible for multiple processes to collaborate on the same data via the filesystem. This is why Firefox for Android still uses the SQLite implementation of `nsILoginManagerStorage`.
```
ncalexan: Again, explain your conclusion: "... still uses the SQLite implementation of ..., when it would be better (reasons?) to use an Android-native storage solution."

rnewman:
Review comments addressed. r?, @ncalexan and @bgrins?

bgrins (approved):
Thanks for writing this up! It summarizes the XULStore implementation correctly AIUI.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- There is a tension between durability (that is: writes that complete are permanent) and performance. We typically choose not to flush the file immediately and synchronously after every change, but not doing so leaves a window in which a crash would cause data loss. By default, that window is 1.5 seconds, plus the time needed to write and atomically switch the files. This forces careful consumers to manually flush if they want their writes to stick, which is a bad pattern.
+- In-memory objects lack the sophistication of most databases. This leads to front-end features building their own [simple query engines](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/storage-json.js#296) for [finding records by linear search](https://dxr.mozilla.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#1081).
+- Similarly, these stores must reinvent their own:
+  - Versioning and upgrade logic.
+  - Validation and schema checking.
+  - Concurrent access patterns.
+  - Write coalescing/async update patterns.
+  - Tooling, if any.
+  - Backup, if any.
+  - [Compression](https://bugzilla.mozilla.org/show_bug.cgi?id=1304389#c45), if any.
```
mnoorenberghe, line 57: Compression is built-in: https://dxr.mozilla.org/mozilla-central/rev/580d833df9c44acec686a9fb88b5f27e9d29f68d/toolkit/modules/JSONFile.jsm#92-93

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- There is a tension between durability (that is: writes that complete are permanent) and performance. We typically choose not to flush the file immediately and synchronously after every change, but not doing so leaves a window in which a crash would cause data loss. By default, that window is 1.5 seconds, plus the time needed to write and atomically switch the files. This forces careful consumers to manually flush if they want their writes to stick, which is a bad pattern.
+- In-memory objects lack the sophistication of most databases. This leads to front-end features building their own [simple query engines](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/storage-json.js#296) for [finding records by linear search](https://dxr.mozilla.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#1081).
+- Similarly, these stores must reinvent their own:
+  - Versioning and upgrade logic.
+  - Validation and schema checking.
+  - Concurrent access patterns.
+  - Write coalescing/async update patterns.
+  - Tooling, if any.
+  - Backup, if any.
```
mnoorenberghe, line 56: Basic backup and recovery (e.g. for malformed JSON due to I/O failure) can be built-in: https://bugzilla.mozilla.org/show_bug.cgi?id=1309501
That could also include recovery after failed schema validation.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+    > consistent. 
+     
+    This can cause jank.
+- Synchronization of data stored in this way takes some careful thought. Syncs typically take hundreds of milliseconds or more, and involve asynchronous network operations, which makes exclusive synchronous access to the in-memory data between reads and writes infeasible.
+- There is a tension between durability (that is: writes that complete are permanent) and performance. We typically choose not to flush the file immediately and synchronously after every change, but not doing so leaves a window in which a crash would cause data loss. By default, that window is 1.5 seconds, plus the time needed to write and atomically switch the files. This forces careful consumers to manually flush if they want their writes to stick, which is a bad pattern.
+- In-memory objects lack the sophistication of most databases. This leads to front-end features building their own [simple query engines](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/storage-json.js#296) for [finding records by linear search](https://dxr.mozilla.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#1081).
```
mnoorenberghe, line 49: IIRC, the passwordmgr example pre-dates the use of JSON and is from the time of using a flat .txt file for logins btw.

I would argue that the simplicity of JSONFile.jsm is one of its advantages… developers don't need to learn any new language/syntax as they just use regular JS features.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+    
+    JSONFile.jsm advises that callers do all of their work synchronously on the main thread to avoid the possibility of concurrent readers or partial writes:
+    
+    > The raw data should be manipulated synchronously, without waiting for the
+    > event loop or for promise resolution, so that the saved file is always
+    > consistent. 
+     
+    This can cause jank.
+- Synchronization of data stored in this way takes some careful thought. Syncs typically take hundreds of milliseconds or more, and involve asynchronous network operations, which makes exclusive synchronous access to the in-memory data between reads and writes infeasible.
+- There is a tension between durability (that is: writes that complete are permanent) and performance. We typically choose not to flush the file immediately and synchronously after every change, but not doing so leaves a window in which a crash would cause data loss. By default, that window is 1.5 seconds, plus the time needed to write and atomically switch the files. This forces careful consumers to manually flush if they want their writes to stick, which is a bad pattern.
```
mnoorenberghe, line 48: The alternative would be always flushing which would be worse for performance so I think it makes sense to give the consumer the choice.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- In-memory objects lack the sophistication of most databases. This leads to front-end features building their own [simple query engines](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/storage-json.js#296) for [finding records by linear search](https://dxr.mozilla.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#1081).
+- Similarly, these stores must reinvent their own:
+  - Versioning and upgrade logic.
+  - Validation and schema checking.
```
mnoorenberghe, line 52: Validation and schema checking is something that could be implemented by JSONFile

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+
+Advantages of this approach:
+- It's relatively simple, with serialization and deserialization being the only steps.
+- Because the file is rewritten frequently, there is some hard-to-measure resilience to gradual file corruption.
+- The file on disk is human-readable and editable, which is convenient for development, test, and support.
+- The in-memory data can be queried and manipulated synchronously.
+- Development is 'natural' for front-end developers — get things working with only in-memory data, then add persistence on top.
+
+Disadvantages:
+- Versioning is often an afterthought: engineers rarely think to version in-memory data formats.
```
mnoorenberghe, line 28: I'm not sure this is a problem that JSONFile causes and it's easy to add versioning when making the first format change by defaulting un-versioned records to version 1.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- Change tracking is almost always an afterthought, thanks to the 'natural' development process: adding syncing later becomes difficult.
```
mnoorenberghe, line 29: A counter-argument is autofill which worked with the Sync team from the beginning and used JSONFile successfully. IIUC, the autofill sync engines (address and credit card) are quite generic and could be easily adapted to support other JSONFile-based stores.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- Users can edit the file on disk, which can result in persistent state that was never created by the component's own logic.
```
mnoorenberghe, line 30: This is true of any file-based storage format though. I guess you want to say that user are more easily able to edit a text-based format over a binary format.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- Is only ever directly read or written by a single thread.
+- Needs to be kept entirely in memory for reads.
+- Has a straightforward JSON encoding (encoding binary blobs, such as images, in base64 is relatively inefficient and typically avoided).
+- Has limited durability requirements.
+
+Advantages of this approach:
+- It's relatively simple, with serialization and deserialization being the only steps.
+- Because the file is rewritten frequently, there is some hard-to-measure resilience to gradual file corruption.
+- The file on disk is human-readable and editable, which is convenient for development, test, and support.
+- The in-memory data can be queried and manipulated synchronously.
```
mnoorenberghe, line 24: Another advantage is that the data can be queried quickly since it's in-memory

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- The entire file typically needs to be read into memory to be used, which increases memory footprint and can impact startup time.
+- Writes at shutdown harm the user experience and don't happen during a crash.
+- In order to achieve atomic file writes, the entire file contents briefly exist twice on disk, which could be problematic on mobile platforms for large data sets.
+- The only approach to cross-process use is duplication of all or part of the in-memory data, which can increase memory footprint. It's not possible for multiple processes to collaborate on the same data via the filesystem. This is why Firefox for Android still uses the SQLite implementation of `nsILoginManagerStorage`: it allows the Android-side `ContentProvider` code to read and write saved logins without worrying about Gecko, which nominally owns the backing storage, but has a shorter lifetime than the enclosing Android code. Using desktop's JSON-backed storage on Android would require launching Gecko on each sync, using messaging for safe access to the file contents.
```
mnoorenberghe, line 35: As a side-note, JSONFile was born out of the (unfortunate) war-on-SQLite days which should have been the war-on-main-thread-synchronous-SQLite days. In retrospect, we shouldn't have moved to JSON simply due to the maintenance cost of having two backends.

text/0012-jsonfile.md
```diff
@@ -0,0 +1,60 @@
...
+- There is a tension between durability (that is: writes that complete are permanent) and performance. We typically choose not to flush the file immediately and synchronously after every change, but not doing so leaves a window in which a crash would cause data loss. By default, that window is 1.5 seconds, plus the time needed to write and atomically switch the files. This forces careful consumers to manually flush if they want their writes to stick, which is a bad pattern.
+- In-memory objects lack the sophistication of most databases. This leads to front-end features building their own [simple query engines](https://dxr.mozilla.org/mozilla-central/source/toolkit/components/passwordmgr/storage-json.js#296) for [finding records by linear search](https://dxr.mozilla.org/mozilla-central/source/browser/extensions/formautofill/FormAutofillStorage.jsm#1081).
+- Similarly, these stores must reinvent their own:
+  - Versioning and upgrade logic.
+  - Validation and schema checking.
+  - Concurrent access patterns.
+  - Write coalescing/async update patterns.
+  - Tooling, if any.
```
mnoorenberghe, line 55: An advantage of JSON is that there is a lot of existing tooling e.g. `jq`

mnoorenberghe (commented):
Hey Richard, 
Here are some comments since I'm quite familiar with the module and major consumers.

text/0012-jsonfile.md
rnewman: I agree that a choice is better than no choice in JSONFile's case.

The main alternative, IMO, is to not routinely force consumers to trade performance versus durability — stores exist that don't require an entire file to be written for a single small change.

text/0012-jsonfile.md
rnewman: Simplicity is the first bullet item in "Advantages of this approach", above; I agree. (I don't really agree that using JS language features is better than having features provided by a storage system, but it's certainly simple!)

text/0012-jsonfile.md
rnewman: I've added a note that consumers could choose to 'upstream' their implementations of these features to JSONFile itself, as was done with compression, should that bug ever be implemented.

text/0012-jsonfile.md
rnewman: I think it's very much caused by solutions like JSONFile: any solution that doesn't itself include versioning, and works by storing "regular JS features", encourages developers to just do what they do for in-memory values, which is nothing at all.

Defaulting doesn't work well for downgrades: the v1 code will (quietly or loudly) misbehave when it loads v2 data, because it doesn't know where to check for the version, or even that it should.

text/0012-jsonfile.md
rnewman: Yep, it's possible to do!

text/0012-jsonfile.md
rnewman: Yep. I'll tweak the wording.

text/0012-jsonfile.md
rnewman: Aye, lessons learned over time!

PR #61, opened 2018-03-14
Add note about lightweight reviews

joewalker:


mykmelez (approved):
This is eminently reasonable.

[blind-review](https://github.com/zombie/blind-reviews "Review performed in blind mode")

PR #64, opened 2018-03-26
Copy mission statement and aspirations from OKRs

joewalker:
Don't merge just yet in case we tweak our aspirations for 2018

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,27 +5,6 @@
 
 Change Mozilla. Investigate big technical challenges and produce engineering programs to address them.
 
-## 2018 Aspirations
-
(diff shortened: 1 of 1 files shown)


mykmelez:
@joewalker I don't think we've tweaked our aspirations since you made this pull request. Is it ok to merge this now?

(Of course, we haven't achieved all our aspirations. But this document is presumably intended to reflect our aspirations rather than our achievements.)

hildjj:
Let's have a discussion about this, making sure that the text adequately represents what the boundaries are going to be.

mykmelez (approved):
Looks good, r+.

PR #65, opened 2018-04-03
Update Android post to remove now redundant steps on building your ow…

fluffyemily:
…n standalone NDKs

This step is now no longer valid - Rustup can find and install these architectures all by itself. Makes cross compiling for Android much simpler.


rnewman (approved):
If it works, it's good enough for me!

PR #70, opened 2018-05-23
WIP - JNA post with no extra diagrams

fluffyemily:


experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+layout: text
+---
+
+# Building and Deploying a Rust library on Android
+
+In our previous post we built a Rust library, and linked to it on [Android using JNI](https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html). This post builds on that by showing how to build the same app using JNA instead of JNI.
+
+# Why JNA?
+
+With JNI, for every `extern "C"` function that we created for our FFI, we had to create a mirror function in Rust for JNI to access. This involved some complicated function name construction and forced us to duplicate large chunks of code. JNA does away with all of that. You create a single interface that declares your FFI functions in Java and that's about it. It removes a large chunk of the headache and it is what we have moved to using here at Mozilla for our Android FFI.
```
grigoryk, line 12: Might be worth mentioning that it does so at some perf cost - which, anecdotally, could be significant depending on the flow of stuff over this channel.

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+mkdir jniLibs/arm64
+mkdir jniLibs/armeabi
+mkdir jniLibs/x86
+
+ln -s <project_path>/greetings/cargo/target/aarch64-linux-android/release/libgreetings.so jniLibs/arm64/libgreetings.so
+ln -s <project_path>/greetings/cargo/target/armv7-linux-androideabi/release/libgreetings.so jniLibs/armeabi/libgreetings.so
+ln -s <project_path>/greetings/cargo/target/i686-linux-android/release/libgreetings.so jniLibs/x86/libgreetings.so
+```
+
+After we've linked out Rust library, we also need to add the `libjnidispatch.so` library for each architecture in the same directories. Right now, I am not sure where we get this from but I have links to pre-built versions of them here:
```
grigoryk, line 120: In practice, I think, this might be linked-in via some CI process artifact. I think this is fine for the purposes of a basic tutorial though.

grigoryk (approved):

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+Now we're all set up and we're ready to start. Let's create the `lib` directory. If you've already created a Rust project from following either or our previous posts, you don't need to do it again.
+
+```
+cargo new cargo
+mkdir android
+```
+
+`cargo new cargo` sets up a brand new Rust project with its default files and directories in a directory called `rust`. The name of the directory is not important. In this directory is a file called `Cargo.toml`, which is the package manager descriptor file, and there is be a subdirectory, `src`, which contains a file called `lib.rs`. This will contain the Rust code that we will be executing.
+
+Our Rust project here is a super simple Hello World library. It contains a function `rust_greeting` that takes a string argument and return a greeting including that argument. Therefore, if the argument is "world", the returned string is "Hello world".
```
mykmelez: Nit: and return -> and returns

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+# Building our library
+
+Now we're all set up and we're ready to start. Let's create the `lib` directory. If you've already created a Rust project from following either or our previous posts, you don't need to do it again.
+
+```
+cargo new cargo
+mkdir android
+```
+
+`cargo new cargo` sets up a brand new Rust project with its default files and directories in a directory called `rust`. The name of the directory is not important. In this directory is a file called `Cargo.toml`, which is the package manager descriptor file, and there is be a subdirectory, `src`, which contains a file called `lib.rs`. This will contain the Rust code that we will be executing.
```
mykmelez: Nit: there is be -> there is

Also, the directory will be called *cargo*, not *rust*, after invoking `cargo new cargo`.

Finally, it seems like `cargo new cargo` uses a binary template rather than a library template by default. To get cargo to create src/lib.rs, you need to specify the `--lib` argument.

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+In our previous post we built a Rust library, and linked to it on [Android using JNI](https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html). This post builds on that by showing how to build the same app using JNA instead of JNI.
+
+# Why JNA?
+
+With JNI, for every `extern "C"` function that we created for our FFI, we had to create a mirror function in Rust for JNI to access. This involved some complicated function name construction and forced us to duplicate large chunks of code. JNA does away with all of that. You create a single interface that declares your FFI functions in Java and that's about it. It removes a large chunk of the headache and it is what we have moved to using here at Mozilla for our Android FFI.
+
+# Getting set up.
+
+Follow the instructions in [our previous post](https://mozilla.github.io/firefox-browser-architecture/experiments/2017-09-21-rust-on-android.html) for building and creating your own standalone toolchains and linking them to Rustup. Once you have your environment set up, head back here for what to do next.
```
mykmelez: Later in this post there's a reference to a _greetings_ directory that is the parent of the _android_ directory, but this post doesn't describe creating it.

The previous post does describe doing so, as a precursor to creating the standalone toolchains. But folks that don't need to do that (either because they're on a platform where it isn't necessary or because they've already done it globally) might not follow the instructions in the previous blog post.

So I recommend including the instructions for creating the _greetings_ directory here as well, i.e.:

>First create a directory for our project.
>
>```
>mkdir greetings
>cd greetings
>```

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+cargo build --target i686-linux-android --release
+```
+
+Now let's create our Android project.
+
+Open Android Studio and select `Start a New Android Project` from the options.
+
+![Create new Android project](assets/2017-09-21-rust-on-android/new-project.png)
+
+On the next screen, type a project name of `Greetings` into the `Application name` field, choose your `Company domain` and select the `android` directory we created earlier as the `Project location`. This will create your Android project inside `greetings/android/Greetings`. Click `Next`.
```
mykmelez: Strangely, when I follow these instructions, Android Studio doesn't append "Greetings" to the path:

screen shot 2018-05-23 at 15 18 25

Although it does do so if I don't change the `Project location`. However, in that case, Android Studio wants to put the project in the last directory I used to create a new project rather than the directory I want to use:

screen shot 2018-05-23 at 15 20 34

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+Now we will be asked to choose a starting activity. Select the `Empty Activity` option and click `Next`.
+
+![Choose starting activity](assets/2017-09-21-rust-on-android/add-activity.png)
+
+Name your Activity and layout on the following screen, calling the activity `GreetingsActivity` and the layout `activity_greetings`. Click `Finish`.
+
+![Name activity](assets/2017-09-21-rust-on-android/customize-activity.png)
+
+We now have a build Rust library and an Android project. We now need to link the two. First, we have to add JNA to our project. Open your applications `build.gradle` file and add the following to the `dependencies` block. When prompted, resync gradle.
```
mykmelez: Nit: applications -> application's

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+In order to build our library, we need to tell the compiler what type of a library it should produce. You can specify this in the `Cargo.toml` file's `[lib]` section:
+
+```
+[lib]
+crate-type = ["dylib"]
+```
+
+We are now ready to build our libraries. Unlike with iOS, there is no handy universal Android library that we can make so we have to create one for each of our target architectures. We can then create symlinks to them from the Android project. You will need to use absolute paths to your libraries here, not relative ones, otherwise Android Studio will not be able to follow the link. Navigate to your `cargo` directory and run the following commands:
```
mykmelez: The statement about needing to use absolute paths would make more sense down by the command snippet that creates the symlinks, as that's where users will be cutting, pasting, and editing to create the symlinks.

experiments/2018-05-23-rust-on-android-JNA.md
mykmelez: JNA provides pre-built versions in platform-specific JAR archives here:

https://github.com/java-native-access/jna/tree/master/dist

And here:

https://github.com/java-native-access/jna/tree/master/lib/native

Which is more than a bit confusing! And [Android Development Environment](https://github.com/java-native-access/jna/blob/master/www/AndroidDevelopmentEnvironment.md) doesn't mention the platform-specific archives. I found them by following links in issue comments.

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+public interface JNA extends Library {
+    String JNA_LIBRARY_NAME = "greetings";
+    NativeLibrary JNA_NATIVE_LIB = NativeLibrary.getInstance(JNA_LIBRARY_NAME);
+
+    JNA INSTANCE = (JNA) Native.loadLibrary(JNA_LIBRARY_NAME, JNA.class);
+}
+```
+
+Here we are declaring that our library is called `greetings`. You will remember that our library is actually called `libgreetings.so`. When Java loads native libraries it will automatically prepend `lib` and append `.so` onto whatevery you call the library during your import. If you were to enter `libgreetings.so` here, the library loader will look for a file called `liblibgreetings.so.so` so be careful that you don't rename your library to something else.
```
mykmelez: Nit: whatevery -> whatever

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+In is inside this interface that we will declare the headers that will link to our Rust library. We have only one external function to declare here, the `rust_greeting` function. Add the following interface declaration to your `JNA` interface.
```
mykmelez: Nit: In -> It

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+```
+String rust_greeting(String pattern);
+```
+
+That's it. We can now create some code that will call it.
+
+
+As with iOS, we're going to create a wrapper class to wrap the JNA binding and ensure that we have a more ergonmic Java API. In the project explorer on the left hand side of the studio window, ensure that `app > java > <domain>.greetings` is highlighted then go to `File > New > Java Class`. Name your class `RustGreetings` and click `OK`.
```
mykmelez: Nit: ergonmic -> ergonomic

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+
+```java
+   static {
+        System.loadLibrary("greetings");
+    }
+```
+
+This looks for a library called `greetings.so` inside the jniLibs directory and picks the correct one for the current architecture.
+
+Open `res/layout/activity-greetings.xml`. In the `Component Tree` panel, highlight the `TextField` and open the `Properties` panel. Change the `ID` in the `Properties` panel to `greetingField`. This is how we are going to refer to it from our Activity.
```
mykmelez: It looks like `TextField` is a `TextView` and `Properties` is `Attributes` on my Android Studio (version 3.1.2 on macOS). Perhaps a platform-specific difference, or changes in a recent Android Studio update?

screen shot 2018-05-23 at 15 52 54

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,198 @@
...
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_greetings);
+
+        RustGreetings g = new RustGreetings();
+        String r = g.sayHello("world");
+        ((TextView)findViewById(R.id.greetingField)).setText(r);
+    }
+```
+
+Build and run the app. If this is your first time in Android Studio, you may need to set up a simulator. When choosing/creating your simulator pick one with API 26. When the app starts, `Hello world` will be printed on your screen.
```
mykmelez: It'd be helpful to explain why the emulator should use API 26 (and the app should require API 26). API 26 isn't available for ARM images, and I've been able to test on an ARM emulator successfully with API 25. The x86 emulator is much faster, of course, and API 26 is available for that. I'm just curious about the reasoning behind the choice of API level, and I suspect your other readers will be as well!

mykmelez (commented):
Overall looks pretty good and is more straightforward than using JNI! I noticed some typos as well as a few more substantive issues. See the code comments for details.

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: Good idea. I'll mention a little bit here and link to https://github.com/java-native-access/jna/blob/master/www/FrequentlyAskedQuestions.md#how-does-jna-performance-compare-to-custom-jni which goes into more detail.

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: This is certainly one bit of the post that I am going to make sure is better before I move this from `WIP`.

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: I think I might have missing this as I was moving stuff around in the post. Thanks for spotting.

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: This must have been a change to Android Studio since I wrote the first post. I will update

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: Yup that is an update to Android Studio. Good spot.

experiments/2018-05-23-rust-on-android-JNA.md
fluffyemily: I must've typo'd somewhere in the original post and it went through unnoticed. Looking back at my notes, it should be API 25.

experiments/2018-05-23-rust-on-android-JNA.md
```diff
@@ -0,0 +1,205 @@
...
+
+`extern` tells the Rust compiler that this function will be called from outside of Rust and to therefore ensure that it is compiled using C calling conventions.
+
+The string that `rust_greeting` accepts is a pointer to a C char array. We have to then convert the string from a C string to a Rust `str`. First we create a `CStr` object from the pointer. We then convert it to a `str` and check the result. If an error has occurred, then no arg was provided and we substitute `there`, otherwise we use the value of the provided string. We then append the provided string on the end of our greeting string to create our return string. The return string is then converted into a `CString` and passed back into C code.
+
+In order to build our library, we need to tell the compiler what type of a library it should produce. You can specify this in the `Cargo.toml` file's `[lib]` section:
+
+```
+[lib]
+crate-type = ["dylib"]
```
vladikoff, line 68: @fluffyemily we had to use `cdylib` in our fxa-rust-client. In projects that have `*.so` dependencies of dependencies with `dylib` you would get `libstd-[hex_value].so`(Rust stdlib?) not found. 

In addition the generated size of `cdylib` is smaller.

Ref: https://stackoverflow.com/a/44175499/216672

vladikoff: I'm not sure what is the best way to explain the differences and usages of the above in this blog post.

salif (approved):

Issue #78, opened 2019-04-01
"Own" the Project Specific Etiquette from the CODE_OF_CONDUCT


joewalker:
The [raw source to the CODE_OF_CONDUCT](https://raw.githubusercontent.com/mozilla/firefox-browser-architecture/master/CODE_OF_CONDUCT.md) contains:

``` html
<!--
## Project Specific Etiquette

In some cases, there will be additional project etiquette i.e.: (https://bugzilla.mozilla.org/page.cgi?id=etiquette.html).
Please update for your project.
-->
```

We should update that for this project.

At the same time, we should link from the README

mykmelez:
Several of us discussed this today and concluded that this project doesn't need project-specific etiquette, as it isn't active enough to have needs beyond Mozilla's standard community participation guidelines.

@joewalker I'll leave this open in case you disagree or want to discuss further (and also to track referencing the Code of Conduct in the README).

PR #80, opened 2019-04-08
Update the README to include the text from the Code of Conduct

joewalker:
Rather than using a link to the Code of Conduct, I just included it verbatim as it's short.

This solution isn't DRY, but forcing users to visit a link for a couple of sentences which are themselves just a link, didn't seem very friendly.


mykmelez (approved):

joewalker:
This closes #78.

mykmelez:
While failing to find "About Firefox" in Firefox's Help menu on Mac this morning, I was reminded that DRY is overrated for human interfaces, since people don't always know where to find the thing they're looking for, and providing it to them in the places they tend to look, even if it duplicates interface elements, is more humane than forcing them to look in the right place or merely pointing them in the right direction.

All of which is to say that I think your intuitions to include the CoC in the README, and to copy it rather than linking to it, are both sound and wise!

